PR #19, opened 2024-06-09
Feature-gate higher order registers

### LINKED ISSUE #18: Limit VM registers to 16 for RV32E
Right now the VM implements 32 registers. It can run RV32E programs, which only use 16 registers, just fine, but we should probably not let the VM use the higher 16 registers when running in "E" mode. An RV32E program that tried to use a register >= 16 should produce an error.

lrettig:
Closes #18

Note: lots of tests, including those in core/src/runtime/mod.rs and core/src/runtime/utils.rs, currently utilize higher registers. These will need to be rewritten.


lrettig:
#18 is still relevant but this PR is stale

Issue #13, opened 2024-06-05
Reproducible builds


lrettig:

schinzelh:
What's your preference here? Something like Guix?

lrettig:
It's not something I know a lot about, and I don't have a strong preference. Ideally it could be done using just Docker (see [Telegram example](https://core.telegram.org/reproducible-builds)), but this may not be possible with Rust. My understanding is that Rust support for reproducible builds is still a [work in progress](https://github.com/rust-lang/rust/labels/A-reproducibility). Welcome any contributions here.

lrettig:
I tried doing some work on this in #49, just to ensure that the `examples` and `tests` are built deterministically. I tried obvious things like adding the relevant `Cargo.lock` files, ensuring the toolchain matches, etc., to no avail. I saved the work in #51. Let's revisit this later.

PR #51, opened 2024-07-22
Make examples and tests builds reproducible

lrettig:


lrettig:
Note: we'll have to use dockerized build for this. See https://docs.succinct.xyz/writing-programs/compiling.html#production-builds.

lrettig:
#13 is still relevant but this PR is stale

PR #57, opened 2024-07-23
Misc improvements

lrettig:
One bugfix, and improve recursive host tests


lrettig (commented):
It looks like solana-sdk v2 is currently unreleased: https://github.com/solana-labs/solana/blob/master/CHANGELOG.md.

Shouldn't we stick with the latest stable release? Right now that's v1.17

PR #58, opened 2024-07-23
Add athena book

hasanza:
Adds the Athena Book.


lrettig:
@hasanza I assumed you'd create the book here by cloning https://github.com/succinctlabs/sp1/tree/main/book, but this looks very different from that. I added some scripts to lint and build the book but they're not working. What did you use as your starting point here?

hasanza:
@lrettig I created a book from scratch using `mdbook`. I thought that would be better and we could bring in material as needed.

hasanza:
@lrettig I investigated the workflow file and it appears that the directory in which the linting, building, and archiving workflows are run is the repository root. I fixed this by adding the correct working directory for the work flows and now everything appears to work properly.

hasanza:
Removed the build folder and added it to git ignore in latest commit.

Issue #61, opened 2024-07-23
Consider replacing `Arc` with `Rc`


lrettig:
The host provider is wrapped in `Arc` but we're not doing any concurrency. This leads to a lot of clippy warnings, which I've manually disabled for now:

```
error: usage of an `Arc` that is not `Send` and `Sync`
   --> interface/src/lib.rs:483:18
    |
483 |       let host = Arc::new(RefCell::new(provider));
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `Arc<RefCell<HostProvider<MockHost<'_>>>>` is not `Send` and `Sync` as `RefCell<HostProvider<MockHost<'_>>>` is neither `Send` nor `Sync`
    = help: if the `Arc` will not used be across threads replace it with an `Rc`
    = help: otherwise make `RefCell<HostProvider<MockHost<'_>>>` `Send` and `Sync` or consider a wrapper type such as `Mutex`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#arc_with_non_send_sync
    = note: `-D clippy::arc-with-non-send-sync` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::arc_with_non_send_sync)]`
```

Decide if we actually need Arc here, or consider switching to Rc.

PR #74, opened 2024-08-01
Add host function SDK and procedural macros

### LINKED ISSUE #54: User-facing SDK
We need a simple developer-facing SDK that hides the unsafe code and other FFI conventions from the smart contract developer. It should use friendlier types and errors, etc., and allow the developer to write purely idiomatic Rust without worrying about what's going on under the hood.

e.g., from this:

https://github.com/athenavm/athena/blob/1224a41e1fda05b7cd59d1c64cf243b71f950244/tests/recursive_call/src/main.rs#L30-L41

to something like this:

```rust
  let return_value = athena_vm::host_wrapper::call(address, val, amount);
```

### LINKED ISSUE #71: Extract symbol/section table info from ELF
Right now, we read all of the code from the ELF into the instructions array, but we don't save the symbols or section headers. This means we can't look up particular functions. We can only execute the program from the main entrypoint.

https://github.com/athenavm/athena/blob/04a249370fb9ead4d710300469c4d3bf0b716e72/core/src/disassembler/elf.rs#L76-L151

It would be nice if we could call specific functions: e.g., going straight to a `verify` function in a template, without requiring the developer to implement some sort of dispatch table in the entrypoint function. Here's some sample code:

```rust
use elf::{ElfBytes, endian::AnyEndian};
use std::fs::File;
use std::io::Read;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    // Read the ELF file
    let mut file = File::open("sample")?;
    let mut buffer = Vec::new();
    file.read_to_end(&mut buffer)?;

    // Parse the ELF file
    let elf = ElfBytes::<AnyEndian>::minimal_parse(&buffer)?;

    // Function name we want to extract
    let function_name = "add";

    // Find the symbol for our function
    let symbols = elf.symbol_table()?
        .ok_or("No symbol table found")?;
    let function_symbol = symbols.iter().find(|sym| {
        sym.get_name(&elf).map(|name| name == function_name).unwrap_or(false)
    }).ok_or("Function not found")?;

    // Get the section containing the function
    let sections = elf.section_headers()?;
    let text_section = sections.get(function_symbol.st_shndx as usize)
        .ok_or("Section not found")?;

    // Calculate the offset and size of the function within the section
    let function_offset = function_symbol.st_value - text_section.sh_addr;
    let function_size = function_symbol.st_size;

    // Extract the function's machine code
    let section_data = elf.section_data(&text_section)?;
    let function_code = &section_data[function_offset as usize..(function_offset + function_size) as usize];

    println!("Machine code for function '{}':", function_name);
    for (i, byte) in function_code.iter().enumerate() {
        print!("{:02x} ", byte);
        if (i + 1) % 16 == 0 {
            println!();
        }
    }
    println!();

    Ok(())
}
```

### LINKED ISSUE #73: Make functions and methods exportable and externally callable
Related to #71

### LINKED ISSUE #75: Implement `spawn` host function

lrettig:
Several quality of life improvements for writing program code.

Includes a Spacemesh-compatible single sig wallet template with an end to end test. Adds a spawn host function, macros to export template methods, and a symbol table for calling methods on a binary.

Closes #54 
Closes #71
Closes #73
Closes #75


vm/sdk/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+
+pub fn call(address: Address, input: Option<Vec<u8>>, amount: Balance) {
+  let address = address_to_32bit_words(address);
+  let amount = balance_to_32bit_words(amount);
+
+  // for now, require input to be word-aligned
+  // we can pad the input but need to know more about the contents
+  let (input, input_len) = if let Some(input) = input {
+    if (input.len() % 4) != 0 {
+      panic!("input is not byte-aligned");
```
poszu, line 16:
```suggestion
      panic!("input is not 4-byte-aligned");
```

vm/sdk/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
```
poszu, line 17: NIT: you can use `assert!` for that
```suggestion
    assert!(input.len() % 4 == 0, "input is not byte-aligned");
```

vm/sdk/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+    (
+      input
+        .chunks(4)
+        .map(|chunk| {
+          let mut bytes = [0u8; 4];
+          bytes.copy_from_slice(chunk);
+          u32::from_le_bytes(bytes)
+        })
```
poszu, line 25: As nightly is already required anyway, you could use its nighly-only features like `array_chunks`:
```suggestion
      input
        .array_chunks()
        .map(|chunk| { u32::from_le_bytes(*chunk) })
```

Note: you would need to add `#![feature(array_chunks)]` at the top of the file to enable it.

vm/sdk/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+
+pub trait VerifiableTemplate {
+  fn verify(&self, tx: &[u8], signature: &[u8; 64]) -> bool;
+}
+
+pub trait WalletProgram {
+  fn spawn(owner_blob: *const u8);
+  fn send(&self, send_arguments_blob: *const u8, send_arguments_blob_len: usize);
+  fn proxy(&self, destination: Address, args: &[u8]);
+  fn deploy(&self, code: &[u8]);
```
poszu, line 67: What's the use case for it? Is a wallet program expected to run some code when it gets deployed?

core/src/syscall/host.rs
```diff
@@ -207,3 +207,37 @@ where
...
+  fn execute(&self, ctx: &mut SyscallContext<T>, arg1: u32, arg2: u32) -> Option<u32> {
+    // let athena_ctx = ctx
+    //   .rt
+    //   .context
+    //   .as_ref()
+    //   .expect("Missing Athena runtime context");
+
+    let len = arg2;
+    let blob: Vec<u8> = ctx
+      .slice(arg1, len as usize / 4)
```
poszu, line 232: What if `len` isn't a multiple of 4?

core/src/syscall/host.rs
```diff
@@ -207,3 +207,37 @@ where
...
+      .iter()
+      .flat_map(|&num| num.to_le_bytes().to_vec())
+      .collect();
+
+    // get value from host
+    let host = ctx.rt.host.as_mut().expect("Missing host interface");
+    host.borrow_mut().spawn(blob);
```
poszu, line 239: What was the reason to use `Arc<RefCell<...>>`? 

`RefCell` bypasses Rust's borrow checker rules by doing borrow checks in runtime. It will panic if the value is already borrows.

`Arc` is meant to be used in multi-threaded code, but afaict it's only used on a single thread. Wouldn't `Rc` be better suited here?

examples/fibonacci/program/elf/fibonacci-program
poszu, line 1: Why did all elfs change?

core/src/syscall/host.rs
lrettig: As it stands, we require the template to pad the blob. I think this is okay for now.

core/src/syscall/host.rs
lrettig: See #61. We inherited this from upstream. I _think_ the idea here is that, eventually, they'll support parallelization, and we might as well. We could replace this with `Rc` or just leave it for now.

examples/fibonacci/program/elf/fibonacci-program
lrettig: Every time the VM code changes, the binary artifacts change if you recompile them. I was thinking of strategies to deal with this, like maybe adding a git commit hook to prevent these from being added except explicitly, or using GHA to update them automatically once per release.

vm/sdk/src/lib.rs
lrettig: See https://community.spacemesh.io/t/proposed-athena-design/433

lrettig:
@poszu this is ready to be reviewed. The scope is too large, and it's a bit messy, and the tests aren't all passing. I'll clean it up as much as I can. There seems to be a chicken-and-egg problem where the tests are trying to install the Athena toolchain, which tries to build Athena, which fails because the toolchain isn't installed. I'm not sure yet where this was introduced.

vm/sdk/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+        .as_ptr(),
+      input.len(),
+    )
+  } else {
+    (std::ptr::null(), 0)
+  };
+
+  unsafe {
+    athena_hostfunctions::call(address.as_ptr(), input, input_len, amount.as_ptr());
+  }
```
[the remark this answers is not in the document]
lrettig: Still need to fix this, good catch

vm/declare/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+              if let Pat::Ident(pat_ident) = &*pat_type.pat {
+                let ident = &pat_ident.ident;
+                let ty = &pat_type.ty;
+                Some((quote!(#ident: #ty), quote!(#ident)))
+              } else {
+                None
+              }
+            } else {
+              None
+            }
```
poszu, line 38: This is a bit hard to follow because of many branches. How about simplifying:
```suggestion
            if let FnArg::Typed(pat_type) = arg {
              if let Pat::Ident(pat_ident) = &*pat_type.pat {
                let ident = &pat_ident.ident;
                let ty = &pat_type.ty;
                return Some((quote!(#ident: #ty), quote!(#ident)));
              }
            }
            None
```

vm/entrypoint/src/syscalls/host.rs
```diff
@@ -86,3 +86,28 @@ pub extern "C" fn get_balance(value: *mut u32) {
...
+
+/// Spawn a new instance of a template.
+///
+/// No return value. It either succeeds or reverts.
+/// The host calculates the new program address based on the template,
+/// the state blob, and the principal nonce. The template and nonce are
+/// available in its context and don't need to be passed here. The
+/// owner of the new program is implicit in the blob, which is just a
+/// serialized version of the instantiated template (struct) state.
+#[allow(unused_variables)]
```
poszu, line 98: NIT: instead of disabling linter, you could provide a separate definition for the `not(zkvm)` target.

core/Cargo.toml
```diff
@@ -27,6 +28,9 @@ thiserror = "1.0.60"
 
 [dev-dependencies]
 athena-vm = { path = "../vm/entrypoint" }
+athena-vm-sdk = { path = "../vm/sdk" }
```
poszu, line 31: I'm not sure, is this the right direction of dependencies that core depends on the SDK? As this is only needed for the tests, how about defining the needed things in the test code instead? All it takes to get rid of it is:

```
diff --git a/core/src/runtime/mod.rs b/core/src/runtime/mod.rs
index 0aafae3..decdf23 100644
--- a/core/src/runtime/mod.rs
+++ b/core/src/runtime/mod.rs
@@ -809,7 +809,6 @@ pub mod tests {
     ADDRESS_LENGTH, SOME_COINS,
   };
   use athena_vm::helpers::address_to_32bit_words;
-  use athena_vm_sdk::{Pubkey, PUBKEY_LENGTH};
   use wallet_common::SendArguments;

   use crate::{
@@ -869,7 +868,7 @@ pub mod tests {
     let program = wallet_program();

     // set up host context
-    let owner_pubkey = [1u8; PUBKEY_LENGTH];
+    let owner_pubkey = [1u8; 32];
     let principal = ADDRESS_ALICE;
     let template = ADDRESS_BOB;
     let callee = ADDRESS_CHARLIE;
@@ -901,7 +900,7 @@ pub mod tests {
     #[allow(clippy::arc_with_non_send_sync)]
     let mut runtime = Runtime::new(program, Some(host.clone()), opts, Some(ctx));
     let mut stdin = AthenaStdin::new();
-    stdin.write(&Pubkey(owner_pubkey));
+    stdin.write(&owner_pubkey);
     runtime.write_vecs(&stdin.buffer);

     // make sure the program loaded correctly
```

core/Cargo.toml
```diff
@@ -9,6 +9,7 @@ edition.workspace = true
 
 [dependencies]
 athena-interface = { path = "../interface" }
+athena-vm-sdk = { path = "../vm/sdk" }
```
poszu, line 12: It's not needed
```suggestion
```

ffi/athcon/include/athcon/athcon.h
```diff
@@ -563,6 +577,9 @@ extern "C"
 
     /** Get block hash callback function. */
     athcon_get_block_hash_fn get_block_hash;
+
+    /** Spawn program callback function. */
+    athcon_spawn_fn spawn;
```
poszu, line 582: The Go bindings need to be updated too. The build emits the following warning:
```
# github.com/athenavm/athena/ffi/athcon/bindings/go [github.com/athenavm/athena/ffi/athcon/bindings/go.test]
host.c:19:1: warning: missing initializer for field ‘spawn’ of ‘const struct athcon_host_interface’ [-Wmissing-field-initializers]
   19 | };
      | ^
In file included from host.go:6,
                 from host.c:1:
./../../include/athcon/athcon.h:5[82](https://github.com/athenavm/athena/actions/runs/10428366033/job/28884425903?pr=74#step:4:83):21: note: ‘spawn’ declared here
  582 |     athcon_spawn_fn spawn;
      |                     ^~~~~
```

Perhaps we should add `-Werror` to CFLAGS, wdyt?

vm/sdk/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "athena-vm-sdk"
```
poszu, line 2: TBH it is hard to follow and build a mental model of the whole thing with so many crates. Why are there 2 SDKs (`athena-sdk` and `athena-vm-sdk`)?. Also, there are `athena-lib`, `athena-vmlib`, `athena-vm`... Do you think that some of them could be combined?

The dependency graph (only the workspace):
graph

On the first glance, it seems that `athena-helper` and `athena-builder` could be combined easily.

Also, the `ffitest` crate could probably be structured as [integration tests](https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests) in the `athena-vmlib` create.

core/src/syscall/host.rs
lrettig: Assuming you're okay with this, let me know if not. This is one of many things we can revisit in the future to make the devex better.

vm/entrypoint/src/syscalls/mod.rs
```diff
@@ -30,3 +30,4 @@ pub const HOST_READ: u32 = 0x00_00_00_A0;
 pub const HOST_WRITE: u32 = 0x00_00_00_A1;
 pub const HOST_CALL: u32 = 0x00_00_00_A2;
 pub const HOST_GETBALANCE: u32 = 0x00_00_00_A3;
+pub const HOST_SPAWN: u32 = 0x00_00_00_A4;
```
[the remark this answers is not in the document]
lrettig: Because the code in `vm/entrypoint` is _guest_ code. The guest cannot rely on `core`. There's some duplication as a result but I think it's okay. I suppose we could set up a very simple 'types' crate, something that like, that both host and guest rely on.

interface/src/lib.rs
```diff
@@ -340,11 +424,49 @@ impl<'a> MockHost<'a> {
...
+      principal: *principal,
+      nonce,
+    });
+    address
   }
 
-  fn transfer_balance(&mut self, from: &Address, to: &Address, value: u64) -> StatusCode {
+  pub fn get_spawn_result(&self) -> Option<&SpawnResult> {
+    self.spawn_result.as_ref()
+  }
```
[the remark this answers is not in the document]
lrettig: Because the spawn host call cannot currently return anything. None of the host calls have return values (https://github.com/athenavm/athena/blob/vmsdk/vm/hostfunctions/src/lib.rs). We probably want to change this: #81.

interface/src/lib.rs
```diff
@@ -505,6 +641,32 @@ impl<'a> HostInterface for MockHost<'a> {
...
+
+  fn spawn(&mut self, blob: Vec<u8>) -> Address {
+    // TODO: double-check these semantics and how Spacemesh principal account semantics map to this
+
+    // Extract the necessary values before calling spawn_program
+    let template = self
+      .dynamic_context
+      .as_ref()
+      .expect("missing host context")
+      .template;
```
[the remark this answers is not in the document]
lrettig: Same answer as above, it has to do with the semantics of the host calls. The caller (user template code) doesn't actually know the template code or address. It doesn't need to. The host fetches it from context.

core/Cargo.toml
lrettig: It somehow bothers me to have the same type defined multiple times in multiple places. Maybe this also belongs in `athena-interface`.

vm/sdk/Cargo.toml
lrettig: > Why are there 2 SDKs (athena-sdk and athena-vm-sdk)?

The naming is confusing. One is an SDK used by developers writing Athena programs (guest side). One is an SDK used to fire up the Athena VM and run code (host side).

> there are athena-lib, athena-vmlib, athena-vm... Do you think that some of them could be combined?

`athena-vmlib` is pretty barebones and should be standalone since it's exposed via FFI, and contains all the messy FFI conversion code. I have to take another look at the others. I'm sure we can refactor things to make it cleaner, but we should do that after merging this PR.

The dependency graph is helpful, thanks.

poszu: I created the graph with [cargo-depgraph](https://github.com/jplatte/cargo-depgraph): `cargo depgraph --workspace-only | dot -Tpng > deps.png`

core/Cargo.toml
poszu: It's 1 constant defining the length of the public key. Btw, is the public key length part of the VM? Isn't it _host_ who defines what public keys are and it's "opaque" for the VM?

vm/sdk/src/lib.rs
poszu: Btw, why is nightly used? The code and tests compile on the latest stable.

vm/sdk/src/lib.rs
poszu: TBH, I still don't understand this. Why `deploy()` takes a `&self`? My understanding was that `deploy()` is meant to put the program code on the chain, which can later be `spawn()`-ed to create an instance of it. What am I missing?

poszu: I found out why. The `deploy()` is used to deploy other programs using the funds from this wallet. It's **not** for deploying this program.

poszu (approved):

PR #77, opened 2024-08-02
load the VM library directly from Go

poszu:
The additional indirection in loading the VM shared library via loader.c seems only to complicate things. Instead of only interfacing with the FFI defined in athcon.h, the client code also needs to build the loader.c and talk with it.

I propose removing the C loader code and loading the shared library directly in Go. Doing the same in other languages should be similarly simple.


ffi/athcon/bindings/go/athcon_test.go
```diff
@@ -23,51 +24,45 @@ func init() {
...
-		t.Fatal(err.Error())
-	}
-	defer i.Destroy()
-	if i.Name() != "Athena" {
-		t.Fatalf("name is %s", i.Name())
-	}
-	if i.Version()[0] < '0' || i.Version()[0] > '9' {
-		t.Fatalf("version number is weird: %s", i.Version())
-	}
+	// TODO: would be good if the VM accepted any options to test their behavior
```
lrettig, line 43: we can add some :) trying to keep it simple for now, haha

ffi/athcon/bindings/go/athcon.go
```diff
@@ -95,43 +98,40 @@ const (
...
+	vmName := strings.TrimPrefix(filename, "lib")
+
+	var athcon_create func() *C.struct_athcon_vm
+	purego.RegisterLibFunc(&athcon_create, handle, "athcon_create_"+vmName)
+	vmHandle := athcon_create()
+
+	if vmHandle == nil {
+		return nil, fmt.Errorf("failed to create VM")
+	}
+	return &VM{lib: handle, handle: vmHandle}, nil
```
lrettig, line 122: naming is quite confusing here. `handle` is a file handle on the dylib, and `vmHandle` is a pointer to the VM instance, right? we could probably make this a little clearer.

lrettig (approved):
cool, nice use of purego, and nice to get rid of all the cmake crud

ffi/athcon/bindings/go/athcon.go
poszu: Renamed `handle` to `libHandle`.

PR #78, opened 2024-08-02
Automatically detect arch and feature

### LINKED ISSUE #17: Auto-detection of RV32E
If we're ever going to support both riscv32e and riscv32i, it would be helpful to remove the hardcoded `rv32e` feature flag added in #15 and make this detection automatic instead. See https://github.com/athenavm/athena/issues/1#issuecomment-2155754917.

I don't think it's possible to do this using plain compiler directives. They're not granular enough to distinguish between the two; both show up as `target_arch = riscv32`. However auto-detection could be done in a build script based on the target:

build.rs:
```rust
use std::env;

fn main() {
    // Example: Check an environment variable to determine if we're targeting `riscv32e`
    let target = env::var("TARGET").unwrap_or_default();
    if target.contains("riscv32e") {
        println!("cargo:rustc-cfg=feature=\"riscv32e\"");
    }
}
```

inline:

```rust
#![cfg_attr(feature = "riscv32e", allow(dead_code))]

#[cfg(feature = "riscv32e")]
fn riscv32e_specific_function() {
    // Implementation for riscv32e
}
```

lrettig:
I think we can remove the `rv32e` feature entirely and just auto-detect from the Rust target

Closes #17

### DIFF
--- a/vm/entrypoint/src/lib.rs
+++ b/vm/entrypoint/src/lib.rs
@@ -40,6 +40,7 @@ macro_rules! entrypoint {
 #[cfg(target_os = "zkvm")]
 mod vm {
   use crate::syscalls::syscall_halt;
+  use cfg_if::cfg_if;
 
   use getrandom::{register_custom_getrandom, Error};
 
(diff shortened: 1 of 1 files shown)


lrettig:
#17 is still relevant but this PR is stale

PR #84, opened 2024-08-06
go bindings: refactors around `VM::Execute`

poszu:
Two notable changes around the `VM::Execute` method.

- use `cgo.Handle` for passing handle to the host context instead of using a global map
- removed the extra wrapper for calling `athcon_execute`.

### DIFF
--- a/ffi/athcon/bindings/go/athcon.go
+++ b/ffi/athcon/bindings/go/athcon.go
@@ -187,6 +187,10 @@ func (vm *VM) Execute(
 		value:     athconBytes32(value),
 	}
 	if len(input) > 0 {
+		// Allocate memory for input data in C.
+		// Otherwise, the Go garbage collector may move the data around and
+		// invalidate the pointer passed to the C code.
+		// Without this, the CGO complains `cgo argument has Go pointer to unpinned Go pointer`.
 		cInputData := C.malloc(C.size_t(len(input)))
 		if cInputData == nil {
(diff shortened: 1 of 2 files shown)


ffi/athcon/bindings/go/athcon.go
```diff
@@ -188,21 +164,56 @@ type Result struct {
...
+	gas int64,
+	recipient, sender Address,
+	input []byte,
+	value Bytes32,
+	code []byte,
+) (res Result, err error) {
+	if len(code) == 0 {
+		return res, fmt.Errorf("code is empty")
+	}
+	msg := C.struct_athcon_message{
```
lrettig, line 181: Let's add a comment here that explains why the message doesn't contain the code, despite having a code field. (Need to double check but I think the `code` field on the message is specifically for deploying new code.)

lrettig (approved):

ffi/athcon/bindings/go/athcon.go
poszu: I also was puzzled by this, but ATM I'm not 100% sure when `code` is supposed to be populated. When deploying new code, shouldn't it be placed in `input`?

lrettig: well, we haven't built the deploy opcode yet (#76), we can finalize it when we do that. this design was copied from EVM.

PR #88, opened 2024-08-14
use `std::sync::LazyLock` in place of `lazy_static!`

poszu:
There's an equivalent of `lazy_static!` in the standard library since Rust 1.80:  [`LazyLock`](https://doc.rust-lang.org/stable/std/sync/struct.LazyLock.html).


poszu:
I'm drafting it as it needs updating the Athena toolchain to Rust 1.80.

lrettig (approved):

PR #91, opened 2024-08-15
Fix toolchain build

lrettig:
Fix some small issues in the build toolchain CLI command

This brings over some logic from https://github.com/athenavm/rustc-rv32e-toolchain/pull/5 and https://github.com/athenavm/rustc-rv32e-toolchain/pull/8 to allow building Rust >= 1.80, which is unfortunately duplicative for now. Opened #93 to sort this out.


cli/src/commands/build_toolchain.rs
```diff
@@ -112,32 +116,67 @@ impl BuildToolchainCmd {
...
       .args([
         "-f",
+        "-N",
         "-p1",
         "-i",
         toolchain_dir.join("patches/rust.patch").to_str().unwrap(),
       ])
       .current_dir(&rust_dir)
+      .output()
+      .expect("Failed to run patch command");
```
poszu, line 130: Why panic here? In other places the error is bubbled up with `?`. How about:
```suggestion
      .output()?;
```

poszu (approved):

Issue #62, opened 2024-07-23
Ensure `ffitest` runs using FFI


lrettig:
This test should be importing `athena_vmlib` using the dylib and running the test this way. I had a [build script](https://github.com/athenavm/athena/blob/583051a9896f36dcf63741527037d5597aaaa313/ffi/ffitest/build.rs) in place to this effect. But the build script was causing problems (because it was calling `cargo build` internally, and hanging), so I removed it -- and the test seems to work fine without it. I can't tell if the test is _actually_ using FFI or not, and I'm not sure how to ensure this other than moving it out of the cargo workspace -- but that would cause other problems since it relies on other imports from the workspace.

https://github.com/athenavm/athena/blob/main/ffi/ffitest/src/lib.rs

PR #96, opened 2024-08-22
move integration tests from ffitest to the right crates

poszu:
Moved:
- `test_athcon_create` to `ffi/vmlib/tests/`
- `test_rust_host` to `ffi/athcon/bindings/rust/athcon-client/tests/`

Updated the `athcon-declare` procedural macro to generate a public `athcon_create_vmname` function so it can be imported where needed instead of hacking with _extern_.


ffi/athcon/bindings/rust/athcon-client/src/lib.rs
```diff
@@ -95,10 +90,8 @@ impl AthconVm {
 }
 
 pub fn create() -> AthconVm {
-  unsafe {
-    AthconVm {
-      handle: athcon_create_athenavmwrapper(),
-      host_interface: Box::into_raw(Box::new(host::get_athcon_host_interface())),
-    }
+  AthconVm {
+    handle: athena_vmlib::athcon_create_athenavmwrapper(),
```
lrettig, line 94: This is cleaner, but I'm not 100% sure that it achieves the original purpose, which is to test the FFI functions end to end over FFI. This isn't actually going over FFI, right? I.e., it's one monolithic program.

ffi/athcon/bindings/rust/athcon-client/Cargo.toml
```diff
@@ -8,7 +8,10 @@ license.workspace = true
 edition.workspace = true
 
 [dependencies]
+athena-vmlib = { path = "../../../../vmlib" }
```
lrettig, line 11: This makes me slightly uncomfortable. In my mental map of the codebase, the athcon bindings shouldn't rely on anything external; they're meant to be imported, and to never create a circular dependency. Does this bother you at all?

lrettig (approved):
I'm okay with this, with the two caveats I shared above

lrettig:
This probably closes #62, wdyt?

ffi/athcon/bindings/rust/athcon-client/src/lib.rs
poszu: IMHO, it works the same as before. It's just importing the extern "C" function explicitly instead of hacking around with externs. The following extern declaration:

```rust
extern "C" {
  fn athcon_create_athenavmwrapper() -> *mut ffi::athcon_vm;
}
```

can **only** work when the VM library is linked with the program that wants to call `athcon_create_athenavmwrapper` or `AthconVm::new()` (which calls the former indirectly).

This test will not work because of linker issues (the extern function is declared, but not defined (its code cannot be found):

```rust
#[test]
fn create_vm() {
	AthconVm::new();
}
```

```sh
error: linking with `cc` failed: exit status: 1
  |
  = note: rust-lld: error: undefined symbol: athcon_create_athenavmwrapper
          >>> referenced by lib.rs:23 (ffi/athcon/bindings/rust/athcon-client/src/lib.rs:23)
          >>>               /home/bartosz/workspace/athena/target/debug/deps/athcon_client-80f6a5be35bcb197.a8o110s31e0sc2ia9h8o2pqub.rcgu.o:(athcon_client::AthconVm::new::h9f1953dc58903478)
          collect2: error: ld returned 1 exit status
```

So, 
>I.e., it's one monolithic program

Yes, both ways require the vmlib to be linked during compile time into "one program".


I think the question is how `athcon-client` is supposed to be used. If it's meant to be used with the `vmlib` only, I think this approach is better. However, if it's meant to be used with other implementations of the VM (coming in form of a shared library), then I think it should support opening the libvm.so and calling the `athcon_create_athenavmwrapper` from it (just like the Go bindings do).

Btw, what is the usecase of having Rust bindings in the first place? I feel that, as we have only 1 implementation of the AthenaVM written in Rust, the "user code" should just use the VM code directly, skipping the FFI layer. (using FFI is like Rust -> C -> Rust).

ffi/athcon/bindings/rust/athcon-client/Cargo.toml
poszu: Actually, it does, although I don't understand the part about circular dependencies. I feel the Rust bindings should probably use the same mechanics as the Go ones. I.e. they should open a shared library provided at a given path and call `athcon_create_athenavmwrapper()` from it.

ffi/athcon/bindings/rust/athcon-client/src/lib.rs
lrettig: > I think the question is how athcon-client is supposed to be used. If it's meant to be used with the vmlib only, I think this approach is better. However, if it's meant to be used with other implementations of the VM (coming in form of a shared library), then I think it should support opening the libvm.so and calling the athcon_create_athenavmwrapper from it (just like the Go bindings do).

For now it's only used in the `ffitest` tests. It's also there as a sort of reference code, if someone wants to use Athena with a Rust-based host (e.g., a future rust-spacemesh client).

I have no issue if we want to rewrite this to work like the Go bindings, but it's definitely not high priority.

> Btw, what is the usecase of having Rust bindings in the first place? I feel that, as we have only 1 implementation of the AthenaVM written in Rust, the "user code" should just use the VM code directly, skipping the FFI layer. (using FFI is like Rust -> C -> Rust).

The Rust bindings are strictly for tests (and, as mentioned, reference code). IMHO it's just nice to have a buttoned-up e2e test of the FFI paths self-contained in this repository, so we have some confidence that changes here won't cause downstream issues for FFI clients. Of course we could write something simple in C or even Go to do this, if you think that's better.

poszu: I double-checked: both ways of calling `athcon_create_athenavmwrapper` **don't** actually go over FFI. They are normal Rust function calls. To enforce going through the FFI, the test would need to open a VM shared library and call the function from it.

However, I think it's not important here. These tests aim to test the exported C interface of the vmlib, not the Rust FFI calling, and they achieve this goal.

poszu:
> This probably closes #62, wdyt?

No, it doesn't. These tests don't really go over FFI, they just call the functions exported for the C FFI. But these calls are "pure Rust". However, I'm not sure if there is any value in testing real FFI calls. IMO, these tests aim to test the exported interface, which they do, not to test Rust FFI calls.

PR #97, opened 2024-08-22
simplify AthconVm trait

poszu:
Changes:
- `execute()` takes `host` by reference instead of a pointer - this allows to drop `unsafe` from its definition
- the `host` pointer is checked in the auto-generated `__execute()` function
- dropped `'a` lifetime from `execute` which enforced `code` and `message` to have the same lifetime.


ffi/athcon/bindings/rust/athcon-declare/src/lib.rs
```diff
@@ -336,16 +336,21 @@ fn build_execute_fn(name: &VMName) -> proc_macro2::TokenStream {
...
+          } else {
+            return athcon_vm::ExecutionResult::failure().into();
+          };
 
-          let execution_message: ::athcon_vm::ExecutionMessage = unsafe {
-              msg.as_ref().expect("athcon message is null").into()
+          let execution_message = if let Some(msg) = unsafe { msg.as_ref() } {
+            msg.into()
+          } else {
+            std::process::abort()
```
lrettig, line 353: Can we return a more meaningful error here, as before? Not sure if `abort()` takes an argument, but at the very least we could print to STDERR

ffi/athcon/bindings/rust/athcon-declare/src/lib.rs
```diff
@@ -336,16 +336,21 @@ fn build_execute_fn(name: &VMName) -> proc_macro2::TokenStream {
           use athcon_vm::AthconVm;
 
           // TODO: context is optional in case of the "precompiles" capability
-          if instance.is_null() || msg.is_null() || (code.is_null() && code_size != 0) {
+          if instance.is_null() || (code.is_null() && code_size != 0) {
```
lrettig, line 339: Why not handle `code` the same as `msg`, below?

ffi/vmlib/src/lib.rs
```diff
@@ -32,29 +32,28 @@ impl AthconVm for AthenaVMWrapper {
   }
 
   /// `execute` is the main entrypoint from FFI. It's called from the macro-generated `__athcon_execute` fn.
-  /// Note that we have to pass in raw `host` and `context` pointers here. If we wrap them into the
+  /// Note that we have to pass in raw `context` pointer here. If we wrap it into the
```
lrettig, line 35:
```suggestion
  /// Note that we have to pass in a raw `context` pointer here. If we wrap it into the
```

ffi/vmlib/src/lib.rs
```diff
@@ -32,29 +32,28 @@ impl AthconVm for AthenaVMWrapper {
...
       return AthconExecutionResult::failure();
     }
 
     // Perform the conversion from `ffi::athcon_message` to `AthenaMessage`
     let athena_msg = AthenaMessageWrapper::from(message);
 
     // Unpack the context
-    let host_interface: &ffi::athcon_host_interface = unsafe { &*host };
-    let execution_context = AthconExecutionContext::new(host_interface, context);
+    let execution_context = ExecutionContext::new(host, context);
```
lrettig, line 56: What happens here if `host` is null?

lrettig (changes requested):
This changes the signature of the FFI `execute` function from

```
  unsafe fn execute<'a>(
    &self,
    rev: Revision,
    code: &[u8],
    message: &AthconExecutionMessage,
    host: *const ffi::athcon_host_interface,
    context: *mut ffi::athcon_host_context,
  ) -> AthconExecutionResult {
```

to

```
  fn execute(
    &self,
    rev: Revision,
    code: &[u8],
    message: &AthconExecutionMessage,
    host: &ffi::athcon_host_interface,
    context: *mut ffi::athcon_host_context,
  ) -> AthconExecutionResult {
```

I don't totally understand all of the downstream consequences of moving from a raw pointer to a Rust reference. Is it simply a matter of how Rust treats the data? Does this change in any way how the caller calls this method?

ffi/vmlib/src/lib.rs
poszu: It cannot be `null` here, it's guaranteed because it's a reference, not a pointer. The caller must have a valid reference to an `athcon_host_interface` to call `execute()`.

poszu:
> This changes the signature of the FFI `execute` function from
> 
> ```
>   unsafe fn execute<'a>(
>     &self,
>     rev: Revision,
>     code: &[u8],
>     message: &AthconExecutionMessage,
>     host: *const ffi::athcon_host_interface,
>     context: *mut ffi::athcon_host_context,
>   ) -> AthconExecutionResult {
> ```
> 
> to
> 
> ```
>   fn execute(
>     &self,
>     rev: Revision,
>     code: &[u8],
>     message: &AthconExecutionMessage,
>     host: &ffi::athcon_host_interface,
>     context: *mut ffi::athcon_host_context,
>   ) -> AthconExecutionResult {
> ```
> 
> I don't totally understand all of the downstream consequences of moving from a raw pointer to a Rust reference. Is it simply a matter of how Rust treats the data? Does this change in any way how the caller calls this method?

This method is not part of the FFI. The automatically generated `extern "C" fn __athcon_execute` is. Using a reference instead of a pointer shifts the responsibility to make sure the pointer is valid to the caller (`__athcon_execute`).

Usually, the **unsafe** FFI functions (like `__athcon_execute`) are small wrappers around **safe** Rust code, meant to validate all input parameters and conditions for safe execution in Rust (e.g. transforming unsafe pointers into safe references). The validity of the `host` pointer should be checked at the "FFI glue layer" (in the extern function that is called over FFI) like all the other input arguments already are (the `host` pointer was a weird exception).

ffi/athcon/bindings/rust/athcon-declare/src/lib.rs
poszu: Changed the code to use assertions with an extra message, which is the recommended way for verifying runtime invariants in unsafe code: https://doc.rust-lang.org/std/macro.assert.html#uses

lrettig:
makes sense, thanks for explaining

lrettig (approved):

PR #98, opened 2024-08-22
make Runtime hold mut reference to HostInterface

poszu:
Two notable changes:
- the `Runtime` takes an exclusive reference to the host instead of `Arc<RefCell<HostInterface>`. It makes the code easier to reason about and lets the Rust borrow checker make sure that the borrow rules are not violated at runtime (the RefCell does it in runtime and might panic). I think it makes sense to require that the host outlives the runtime. It also allowed to remove the code of "stealing host" and later rebuilding it from the Arc.
- make the `Runtime` **not** generic over the `HostInterface` but use a dyn trait object instead. It simplifies the code a lot and removes the requirement to specify the host type when creating a runtime without a host (which led to using MockHost, where None host was passed).

### DIFF
--- a/cli/src/commands/execute.rs
+++ b/cli/src/commands/execute.rs
@@ -3,7 +3,6 @@ use anyhow::Result;
 use athena_builder::{build_program, BuildArgs};
 use athena_core::io::AthenaStdin;
 use athena_core::utils::{setup_logger, setup_tracer};
-use athena_interface::MockHost;
 use athena_sdk::ExecutionClient;
 use clap::Parser;
 use std::time::Instant;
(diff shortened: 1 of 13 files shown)


poszu:
:bulb: The Fibonacci template ib https://github.com/athenavm/athena-project-template needs to be updated to pass the CI.

.github/workflows/pr.yml
```diff
@@ -163,7 +163,7 @@ jobs:
 
       - name: Run cargo athena new
         run: |
-          cargo athena new fibonacci
+          cargo athena new fibonacci --version refactor/execution-client
```
lrettig, line 166: remove before merge

lrettig (changes requested):
nice, this is surprisingly clean

.github/workflows/pr.yml
poszu: Removed

lrettig (approved):

PR #99, opened 2024-08-23
[wip] wallet PR improvements

poszu:
Fixed a few bugs:
- passing the program template from program to host via the SPAWN syscall didn't handle situation when the serialized template blob wasn't a multiple of 4 bytes (it missed last bytes). Fixed + added unit tests
- the syscall helpers in vmsdk passed a dangling pointer to the VM because the Vec was immediately deallocated.

Implemented the wallet script to create a mocked host, spawn the wallet and call `send()`.


examples/wallet/program/src/main.rs
```diff
@@ -1,6 +1,5 @@
 //! The Spacemesh standard wallet template.
 #![no_main]
-#![no_std]
```
lrettig, line 3: We need to discuss the implications of this change

vm/sdk/src/spawn.rs
```diff
@@ -0,0 +1,55 @@
...
+      let mut bytes = [0u8; 4];
+      bytes.copy_from_slice(chunk);
+      u32::from_le_bytes(bytes)
+    })
+    .collect::<Vec<u32>>();
+
+  let remainder = chunks.remainder();
+  if !remainder.is_empty() {
+    let mut bytes = [0u8; 4];
+    bytes[..remainder.len()].copy_from_slice(remainder);
```
lrettig, line 21: Just double checking. This always zero-pads to the right, which when combined with LE ends up with zero padding to the left (as in the below test). Is this the intended behavior?

vm/sdk/src/spawn.rs
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  #[test]
+  fn convert_exact_multiple_of_4() {
+    let result = bytes_to_u32_vec([1, 2, 3, 4, 5, 6, 7, 8]);
+    assert_eq!(result, vec![0x04030201, 0x08070605]);
+  }
+
+  #[test]
+  fn covnert_not_a_multiple_of_4() {
```
lrettig, line 45:
```suggestion
  fn convert_not_a_multiple_of_4() {
```

vm/sdk/src/spawn.rs
```diff
@@ -0,0 +1,55 @@
...
+    let result = bytes_to_u32_vec([1, 2, 3, 4, 5, 6, 7]);
+    assert_eq!(result, vec![0x04030201, 0x00070605]);
+  }
+
+  #[test]
+  fn convert_single_byte() {
+    let result = bytes_to_u32_vec([1]);
+    assert_eq!(result, vec![0x01]);
```
lrettig, line 53: This is a bit counterintuitive, worth adding a comment explaining what's going on here

lrettig (changes requested):

examples/wallet/program/src/main.rs
poszu: It's not necessary. I thought it was added by mistake as other examples don't have it. Reverting.

vm/sdk/src/spawn.rs
poszu: I'm not sure I understand. Do you refer to this test?
```
fn convert_not_a_multiple_of_4() {
    let result = bytes_to_u32_vec([1, 2, 3, 4, 5, 6, 7]);
    assert_eq!(result, vec![0x04030201, 0x00070605]);
  }
```

The second word is not left-padded. This is just what hex representation looks like (the lower bytes are on the right). In memory, the underlying data is continuous. When this vec is later written into the "VM memory", it works as expected:


|address | 0 | 1 | 2| 3 |
|---|---|---|---|---|
|0x0 | 0x01 | 0x02| 0x03|0x04|
|0x1 | 0x05 | 0x06| 0x07|0x00|

vm/sdk/src/spawn.rs
poszu: Why is it counterintuitive? I updated the test to show that the u32 is a zero-padded little-endian value and added a doc comment on `bytes_to_u32_vec`.

vm/sdk/src/spawn.rs
lrettig: yes, okay, makes sense now. I think the right way to interpret my comment is, "with LE, we want to zero-pad to the right, i.e., add zeroes to the MSB."

lrettig (approved):

PR #100, opened 2024-08-23
PoC: spawn hostcall returns address on stack

### LINKED ISSUE #82: `spawn` host function should return program address
Spawn host function was implemented (#75) without a return value. The assumption was that the spawn would either fail, in which case the call would revert, or it would succeed, in which case the newly-created program address would be returned in `ExecutionResult.create_address`.

However, it's possible for a single execution thread to call spawn multiple times, and in this case, the address of the programs created by all but the final spawn would be lost. I'm pretty sure we instead need the spawn call to return the newly-created program address. This also simplifies a few things.

Look into the EVM dynamics to understand how it's done there:

- https://www.evm.codes/#f0?fork=cancun
- https://www.evm.codes/#f5?fork=cancun
- https://docs.openzeppelin.com/cli/2.8/deploying-with-create2

This may depend on #81.

poszu:
Closes #82 

Two notable changes here.

First, it adds support for returning the address of spawned program instance from the `SPAWN` host call. The address is shared with the guest program via it's memory (on the stack, but the syscall doesn't care). The pointer to write the result into is shared with the VM via A2 / X12 register. The syscall handler on the VM side writes the result (the address of the spawned program) into the guest program memory pointed to by register X12.

Secondly, adds support for reading program function inputs from IO and converting them into the expected types. The argument type `T` must implement `IntoArgument<T>` for this to work, otherwise it won't compile. The `IntoArgument<T>` is automatically implemented for all `T`s implementing `BorshDeserialize` for ease of use. 
Similarly, the program function result is automatically converted and written to the io using the `IntoResult<T>` trait. 

This allows for adjusting the `trait WalletProgram` so that its functions take arguments and return values. These arguments and values are automatically converted from/to IO thanks to the aforementioned traits.

Thanks to this, the interface of a program can better express its inputs and outputs and be more readable.


jellonek:
lgtm

jellonek (approved):

vm/lib/src/io.rs
```diff
@@ -157,3 +156,28 @@ pub fn hint_slice(buf: &[u8]) {
...
 }
+
+#[derive(Default)]
+pub struct Io {
+  read_remainder: std::collections::VecDeque<u8>,
+}
+
+impl std::io::Read for Io {
+  fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
+    if self.read_remainder.is_empty() {
```
lrettig, line 167: I don't understand the use of `read_remainder` here, why is this necessary? In other words, why not just read directly using `io::read_vec()`? Probably worth an explanatory comment.

vm/entrypoint/src/syscalls/host.rs
```diff
@@ -98,19 +98,24 @@ pub extern "C" fn get_balance(value: *mut u32) {
...
-        in("a1") len,
+        in("a0") blob.as_ptr(),
+        in("a1") bytes_len,
+        in("a2") result.as_mut_ptr(),
     )
   }
 
-  #[cfg(not(target_os = "zkvm"))]
-  unreachable!()
+  unsafe { result.assume_init() }
```
lrettig, line 120: What happens if the spawn fails? Do we get a zero address back? Does it never return? Also worth an explanatory comment

vm/entrypoint/src/syscalls/host.rs
```diff
@@ -98,19 +98,24 @@ pub extern "C" fn get_balance(value: *mut u32) {
...
         in("t0") crate::syscalls::HOST_SPAWN,
-        in("a0") blob,
-        in("a1") len,
+        in("a0") blob.as_ptr(),
+        in("a1") bytes_len,
+        in("a2") result.as_mut_ptr(),
     )
   }
 
-  #[cfg(not(target_os = "zkvm"))]
```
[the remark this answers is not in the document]
lrettig: Your idea is that it does nothing, rather than panicking?

examples/wallet/program/src/main.rs
```diff
@@ -81,15 +76,11 @@ mod test {
...
     let owner: Pubkey = [0u8; 32];
-    Wallet::spawn(owner.as_ptr());
+    Wallet::spawn(owner);
     let send_arguments = SendArguments {
       recipient: [0u8; 24],
       amount: 0,
     };
-    let serialized = to_vec(&send_arguments).unwrap();
-    wallet.send(serialized.as_ptr(), serialized.len());
-    unsafe { athexp_spawn(owner.as_ptr()) };
```
lrettig, line 91: Why remove this spawn call?

core/src/syscall/host.rs
```diff
@@ -202,9 +202,17 @@ impl Syscall for SyscallHostSpawn {
     let vec_words = ctx.slice(address, len_words);
     let blob = vec_u32_to_bytes(vec_words, len as usize);
 
-    // get value from host
     let host = ctx.rt.host.as_deref_mut().expect("Missing host interface");
-    host.spawn(blob);
+    let address = host.spawn(blob);
+
+    let out_addr = ctx
```
lrettig, line 208: Checking memory safety here. Are we sure that a user program cannot sneak in a different memory location here before this write happens? Worth adding a comment here explaining why, it's not immediately obvious.

lrettig (changes requested):
not requesting substantial changes, mostly checking things and suggesting that we add some explanatory comments. good work.

vm/entrypoint/src/syscalls/host.rs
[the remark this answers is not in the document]
poszu: The whole function is compiled only for `target_os = "zkvm"`. It won't compile if one tries to call it from code compiled for another target, instead of panicking at runtime.

vm/lib/src/io.rs
poszu: Because `read()` is called with a small `buf` that doesn't fit the whole result of ``io::read_vec()`. The `remainder` buffer is needed to persist the rest between the calls. This is needed to support deserialization.

For example, say there is a struct which we want to serialize and write to `io` and read and deserialize on the other side. The struct is serialized to a single `Vec<u8>` and written to the io. This vec is later returned "whole" in a single `io::read_vec()`. But the deserializer like borsh might take multiple reads from the provided `reader` to fully deserialize back to the struct.

vm/entrypoint/src/syscalls/host.rs
poszu: It would never return to the guest program. Any failure in a syscall currently ends with panic and killing the whole VM / hosting program.

core/src/syscall/host.rs
poszu: What do you mean by `user program`? Do you mean the guest (the risc-v program)? This one is halt for the entire time of the syscall so it cannot do anything. The `out_addr` and the memory it points to only exist in the context of this `Runtime` instance (this is basically a map in the `Runtime` struct). 

https://github.com/athenavm/athena/blob/8f4dd9ef8839c297b6d9a4238c1c645f829c1e44/core/src/runtime/state.rs#L22

It's not possible to overwrite the host memory this way. Check out the `mw` routine: https://github.com/athenavm/athena/blob/ae54d14c7c8b489240765cafb393d5266665cdc6/core/src/runtime/mod.rs#L222-L237

examples/wallet/program/src/main.rs
poszu: Actually, it was not a mistake, I forgot :) I removed these direct calls to `athexp_` functions because they were wrong and did not compile. These functions don't take any arguments. You can use `cargo expand` to see them: `RUSTUP_TOOLCHAIN=athena cargo expand --target riscv32em-athena-zkvm-elf`:

```rust
pub unsafe extern "C" fn athexp_spawn() {
    athena_vm::program::Function::call_func(
        Wallet::spawn,
        &mut athena_vm::io::Io::default(),
    );
    syscall_halt(0);
}

pub unsafe extern "C" fn athexp_send() {
    athena_vm::program::Method::call_method(
        Wallet::send,
        &mut athena_vm::io::Io::default(),
    );
    syscall_halt(0);
}
```

poszu: You can compile the test with: `RUSTUP_TOOLCHAIN=athena cargo t --no-run --target riscv32em-athena-zkvm-elf`

vm/lib/src/io.rs
lrettig: makes sense. can we add a comment here to explain this? :pray:

examples/wallet/program/src/main.rs
lrettig: okay. agree, it would be nice to have support for this eventually. should we add compilation of these tests to the CI pipeline?

core/src/syscall/host.rs
lrettig: No, I mean a guest program overwriting the memory of another guest program (sorry if I wasn't clear). E.g., imagine a smart contract calling into another smart contract: the first execution frame is paused while the second runs. Depending how this is implemented, they could share memory. (The way it's built today is quite naive/inefficient, an entirely new VM is spun up for the call, so I'm pretty sure they don't share memory, but this might change in the future).

Alternatively: a guest program overwriting its own code. For safety we probably want to maintain the constraint that a program cannot overwrite its own instructions (see https://en.wikipedia.org/wiki/Harvard_architecture).

poszu: > No, I mean a guest program overwriting the memory of another guest program (sorry if I wasn't clear). E.g., imagine a smart contract calling into another smart contract: the first execution frame is paused while the second runs. Depending how this is implemented, they could share memory. (The way it's built today is quite naive/inefficient, an entirely new VM is spun up for the call, so I'm pretty sure they don't share memory, but this might change in the future).

OK, I understand now. You are right, it would be a problem if they shared the memory. But the way the VM `Runtime` is currently implemented, running 2 programs in the same `Runtime` isn't safe - they could easily manipulate the memory of one another by executing store instructions and writing the memory space used by another. The way to separate memory spaces of 2 programs "physically" is to run separate Runtimes.

> Alternatively: a guest program overwriting its own code. For safety we probably want to maintain the constraint that a program cannot overwrite its own instructions (see [en.wikipedia.org/wiki/Harvard_architecture](https://en.wikipedia.org/wiki/Harvard_architecture)).

The program instructions and memory are separated and it's impossible to overwrite the program: https://github.com/athenavm/athena/blob/096857aa5b37e38f466f5c01834da161ce69c06b/core/src/runtime/mod.rs#L38-L46

The program memory is held in `state.memory` separately from the instructions and writing it cannot modify them.

examples/wallet/program/src/main.rs
poszu: Added compiling these tests to the CI.

lrettig (approved):

PR #103, opened 2024-09-11
use stable rust

poszu:
It's unclear why a nightly toolchain was enforced in the rust-toolchain file, but everything compiles on stable.

I also removed filtering in pr.yml because the list of paths was incomplete and there was a risk that some changes wouldn't be tested in CI.


poszu:
Apparently, nightly is required for catching unwinding panics across FFI boundary, which is not supported in stable.

lrettig:
Thanks for investigating - good to know!

poszu:
Reopening it. I changed From impl to Try from for athcon message convertion so it doesn't require catching panics (and unwinding) and we don't require nightly anymore.

jellonek (approved):

PR #107, opened 2024-09-13
PoC gdbstub for debugging guest programs

poszu:
A basic gdbstub that allows for debugging VM guest programs.

Example:

First, run the program with GDB on selected port:
```rust
let elf = include_bytes!("path to program elf");
let stdin = AthenaStdin::new();
let client = ExecutionClient::new();
client.execute_with_gdb(ELF, stdin.clone(), None, None, None, 9001);
```

It will spin up GDB server listening on port 9001 and halt the guest program before executing the first instruction. Note that the Rust VM host program is running, not stopped on any breakpoint etc.

Connect with gdb or IDE debugger:
GDB:
```sh
> gdb
(gdb) file <path to elf with built with symbols>
(gdb) target remote localhost:9001
(gdb) debug normally
```

Vscode:
Edit launch.json configuration and add:
```json
{
    "type": "gdb",
    "request": "attach",
    "name": "Attach to guest program",
    "executable": "<path to elf>",
    "target": ":9001",
    "remote": true,
    "cwd": "${workspaceRoot}",
    "valuesFormatting": "parseText",
    "stopAtConnect": true
},
```

and start debugging with it.


lrettig:
Worth taking a look at this: https://github.com/risc0/risc0/issues/129

lrettig:
I tried adding the recommended configuration to `launch.json` in vscode but I get a bunch of errors, any idea why?

image
image

lrettig:
Found it, it requires this extension: https://marketplace.visualstudio.com/items?itemName=webfreak.debug

lrettig:
A few other notes on getting this to work:

- by default athena builds programs without debug info. You need to manually compile the program in question with a command like this:

```
> RUSTUP_TOOLCHAIN=athena cargo build --target riscv32em-athena-zkvm-elf
```

and use the path to this binary in the `launch.json` config.

- the default Linux gdb doesn't support the riscv architecture. You need gdb from something like https://github.com/riscv-collab/riscv-gnu-toolchain and then you need to set `gdbpath` in `launch.json` as well.
- gdb gets a bit confused about where it is, but if you set a breakpoint in the main code and run "c" (continue) it seems to work. Note that the "Step over", "Step into" etc. commands don't seem to work correctly; you need to use breakpoints.

poszu:
> I tried adding the recommended configuration to `launch.json` in vscode but I get a bunch of errors, any idea why?
> 
> image image

You are probably missing GDB debugging extension in VSCode. Try this one: [Native Debug](https://marketplace.visualstudio.com/items?itemName=webfreak.debug)

lrettig:
Curious why this is a draft, should we aim to merge this? It's clearly useful. IMHO the only thing missing here is documentation.

poszu:
> Curious why this is a draft, should we aim to merge this? It's clearly useful. IMHO the only thing missing here is documentation.

Yeah, documentation is missing but also because it's very basic support and not everything works perfectly. There are some issues with presenting local variables (I'm not sure if this is issue with the integration or vscode though - e.g. gdb itself prints variables nicely from when asked to with `p <variable name>`). I'm curious, what problems with stepping did you have?

Also, it would be great to have an option to build guest programs with _debug_ profile (currently _release_ is hardcoded).

---

> by default athena builds programs without debug info. You need to manually compile the program in question with a command like this:
>
>> RUSTUP_TOOLCHAIN=athena cargo build --target riscv32em-athena-zkvm-elf

Debug info can be easily turned on with placing this in the `Cargo.toml`:
```toml
[profile.release]
debug=true
```

lrettig:
> I'm curious, what problems with stepping did you have?

It didn't work at all. Every time I tried to step over, in, or out, it was the same as "continue."

lrettig (approved):
tweaked README a little, lgtm

PR #108, opened 2024-09-13
deploy host function

### LINKED ISSUE #76: Implement `deploy` host function

poszu:
Closes #76


examples/wallet/script/src/bin/execute.rs
```diff
@@ -116,3 +115,43 @@ fn main() {
...
+    athena_sdk::utils::setup_logger();
+
+    let mut host = MockHost::new_with_context(
+      HostStaticContext::new(ADDRESS_ALICE, 0, ADDRESS_ALICE),
+      HostDynamicContext::new([0u8; 24], ADDRESS_ALICE),
+    );
+    let address = super::spawn(&mut host, Pubkey::default()).unwrap();
+
+    // deploy other contract
+    let code = b"some bad really code".to_vec();
```
lrettig, line 136:
```suggestion
    let code = b"some really bad code".to_vec();
```

this made me laugh :)

examples/wallet/script/src/bin/execute.rs
```diff
@@ -116,3 +115,43 @@ fn main() {
...
+    let mut stdin = AthenaStdin::new();
+    let wallet = host.get_program(&address).unwrap();
```
lrettig, line 138:
```suggestion
    let wallet_state = host.get_program(&address).unwrap();
```

examples/wallet/script/src/bin/execute.rs
```diff
@@ -116,3 +115,43 @@ fn main() {
...
+    stdin.write_slice(wallet);
```
lrettig, line 139:
```suggestion
    stdin.write_slice(wallet_state);
```

interface/src/lib.rs
```diff
@@ -654,6 +659,19 @@ impl<'a> HostInterface for MockHost<'a> {
       static_context.nonce,
     )
   }
+
+  fn deploy(&mut self, code: Vec<u8>) -> Address {
+    // template_address := HASH(template_code)
+    let hash = blake3::hash(&code);
+    let hash_bytes = hash.as_bytes().as_slice();
+    let address = Address::try_from(&hash_bytes[..24]).unwrap();
```
lrettig, line 667:
```suggestion
    let address = Address::try_from(&hash_bytes[..ADDRESS_LENGTH]).unwrap();
```

should update this in at least one more place in this file (where the hasher is used)

interface/src/lib.rs
```diff
@@ -654,6 +659,19 @@ impl<'a> HostInterface for MockHost<'a> {
...
+
+    if self.templates.contains_key(&address) {
+      panic!("template already exists");
```
lrettig, line 670: Panic is fine for now, but let's discuss the right protocol for this. It could:

1. succeed, and just be idempotent (but does it charge gas?)
2. result in a failed tx
3. panic
4. something else?

vm/entrypoint/src/syscalls/host.rs
```diff
@@ -89,7 +89,8 @@ pub extern "C" fn get_balance(value: *mut u32) {
 
 /// Spawn a new instance of a template.
 ///
-/// No return value. It either succeeds or reverts.
+/// It either succeeds or reverts.
+/// Returns the address of the spawned program.
```
lrettig, line 93:
```suggestion
/// In case of success, returns the address of the spawned program.
```

examples/wallet/script/src/bin/execute.rs
```diff
@@ -116,3 +115,43 @@ fn main() {
...
 }
+
+#[cfg(test)]
+mod tests {
+  use athena_interface::{Address, HostDynamicContext, HostStaticContext, MockHost, ADDRESS_ALICE};
+  use athena_sdk::{AthenaStdin, ExecutionClient};
+  use athena_vm_sdk::Pubkey;
+
+  #[test]
+  fn deploying_other_contract() {
```
lrettig, line 126:
```suggestion
  fn deploy_template() {
```

Let's try to stick to the terminology: "deploy" a "template", and "spawn" a "program" ("instance of a template")

ffi/athcon/include/athcon/athcon.h
```diff
@@ -187,6 +187,19 @@ extern "C"
    */
   typedef athcon_address (*athcon_spawn_fn)(struct athcon_host_context *context, const uint8_t *blob, size_t blob_size);
 
+    /**
+   * Deploy program callback function.
+   *
+   * This callback function is used by a VM to deploy a new program template
+   *
+   * @param context   The pointer to the Host execution context.
+   * @param blob      The program's template (serialized code).
```
lrettig, line 196:
```suggestion
   * @param blob      The program's template (bytecode).
```

"serialized code" sounds kinda funny

lrettig (changes requested):
a few nits but overall lgtm!

interface/src/lib.rs
poszu: Error handling across the VM worries me too, and I think this is a much bigger problem than visible here. There are at least 2 kinds of issues, which are both "handled" by panicking currently:
1. failures in the host callbacks (like in this example). There is currently no way for the host to return an error so the only option is to panic.
2. failures in the VM like here, when a program tries to read from IO more than there is: https://github.com/athenavm/athena/blob/7a5b832084c1c42ecd7a72d7631ab803bf8f1b50/core/src/syscall/hint.rs#L22-L30

Both make it possible to crash the VM and the host process with a crafted guest program. I think we should invest some time in thinking about error handling and use panics and unwraps really carefully.

poszu: Now that syscalls are fallible (#110), I changed the `HostInterface::deploy` to return an error.

lrettig (approved):

PR #110, opened 2024-09-17
Don't panic in syscalls

poszu:
Panicking in a system call would result in the VM and host process crashing. We don't want this. Instead, we want to fail program execution.


jellonek (approved):

core/src/syscall/write.rs
```diff
@@ -1,30 +1,37 @@
...
 
-pub struct SyscallWrite;
+/// Write bytes to selected file descriptor.
+/// Supported FDs:
+/// - 1: stdout
+/// - 2: stderr
+/// - 3: public values
+/// - 4: input stream (TODO:poszu check why it writes to the input stream)
+///
+/// FD 1 supports "cycle tracker". TODO(poszu): see if we need this, and add documentation or remove.
```
lrettig, line 15: re: cycle tracker, see https://docs.succinct.xyz/writing-programs/cycle-tracking.html. We can discuss whether we want to maintain this.

re: fd4 and writing to the input stream, I can't find any documentation on this in sp1. My guess is that it's useful for debugging and testing, and in theory could be used for cross-contract calls (inside the same runtime, which we don't support) or recursive proofs (which we don't support yet).

lrettig (changes requested):
One basic question: what's wrong with panicking inside the VM, as we're doing today, and just having the host catch and unwind a failing tx, rather than doing it inside the VM?

poszu:
> One basic question: what's wrong with panicking inside the VM, as we're doing today, and just having the host catch and unwind a failing tx, rather than doing it inside the VM?

Panics are not meant for control flow and handling "normal" failures. In Rust we return `Result` in this case, a "throw/catch" is an antipattern. Panics are fine for situations that should not happen or are inrecoverable, like an invariant that is not upheld. Check out a chapter about this in The Book: https://doc.rust-lang.org/beta/book/ch09-03-to-panic-or-not-to-panic.html
 
> > Panicking in a system call would result in the VM and host process crashing.
> 
> Is it not possible to try and catch an exception or panic using FFI?

Short answer: it is generally not possible, unwinding across FFI is generally an undefined behavior.
Long answer :) It is sometimes possible, an exception can be unwinded and caught through FFI only in Rust (if the caller is a Rust program), and only if the FFI function is declared with "C-unwind" ABI. Catching it in any other language like Go is impossible - this is an undefined behavior (for example unwinding might overwrite the caller's stack). For this reason, I think we should build the vmlib with `panic=abort` to shut down on a panic immediately. Check out: https://doc.rust-lang.org/std/panic/fn.catch_unwind.html, in particular:

>It is currently undefined behavior to unwind from Rust code into foreign code, so this function is particularly useful when Rust is called from another language (normally C). This can run arbitrary Rust code, capturing a panic and allowing a graceful handling of the error.
> 
> It is not recommended to use this function for a general try/catch mechanism. The [Result](https://doc.rust-lang.org/std/thread/type.Result.html) type is more appropriate to use for functions that can fail on a regular basis. Additionally, this function is not guaranteed to catch all panics, see the “Notes” section below.

core/src/syscall/write.rs
poszu: Regarding the cycle tracker, I see a potential problem with it because it lets the contract code control some debugging/profiling functionality in the runtime. I'm unsure if we want to allow doing this by arbitrary programs. Perhaps this could be compiled in under a feature flag, only for builds instrumented for profiling and **not** enabled "in production".

core/src/runtime/syscall.rs
```diff
@@ -127,21 +123,17 @@ impl<'a, 'h> SyscallContext<'a, 'h> {
     self.rt.word(addr)
   }
 
+  pub fn dword(&self, addr: u32) -> u64 {
```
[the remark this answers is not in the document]
poszu: meaning of word, dword, qword and so on depends on architecture I think. Here, I used dword for a `u64` because a `u32` is named "word". So double-word = "2 * u32" = u64.

core/src/runtime/syscall.rs
```diff
@@ -69,13 +70,19 @@ impl SyscallCode {
...
+pub enum Outcome {
+  Result(Option<u32>),
+  Exit(u32),
+}
+
+pub(crate) type SyscallResult = Result<Outcome, StatusCode>;
+
 pub trait Syscall: Send + Sync {
-  /// Execute the syscall and return the resulting value of register a0. `arg1` and `arg2` are the
-  /// values in registers X10 and X11, respectively. While not a hard requirement, the convention
```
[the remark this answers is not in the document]
poszu: No, the "convention" is unchanged - a syscall might return a result (hence the `Result(Option<u32>)` variant of the `Outcome` enum). I only changed the `execute` trait function to:
- allow returning errors (instead of panicking) - hence it returns a `Result<Outcome, StatusCode>`
- avoid doing dynamic dispatch in the Runtime and having special meaning fields in the `SyscallContext` (like the `exit_code`) that were only used in a specific syscall (HALT). Instead, the "you should exit" outcome is just one of the variants in the `Outcome` enum.

core/src/syscall/write.rs
lrettig: works for me. feel free to open an issue to follow up on this.

lrettig (approved):

PR #112, opened 2024-10-01
Add method name to execution message and call syscall

### LINKED ISSUE #111: Allow multiple program entry points over FFI
Internally, we support multiple program entry points via the use of `execute` and `execute_function`:

https://github.com/athenavm/athena/blob/6bdceaab5cf3bf9f44c42ecfda3abb4ec08002f2/sdk/src/lib.rs#L53-L78

When passing in a `message` over FFI/the vmlib, however, there's currently no way to call a particular method:

https://github.com/athenavm/athena/blob/6bdceaab5cf3bf9f44c42ecfda3abb4ec08002f2/runner/src/vm.rs#L61

This is important so the host can call template methods like `spawn` and `spend`.

lrettig:
Closes #111


ffi/athcon/bindings/rust/athcon-vm/src/lib.rs
```diff
@@ -253,6 +261,17 @@ impl<'a> ExecutionContext<'a> {
...
+    let method_name_size = if let Some(method) = method {
+      method.len()
+    } else {
+      0
+    };
+    let method_name = if let Some(method) = method {
+      method.as_ptr()
+    } else {
+      std::ptr::null()
+    };
```
poszu, line 274: NIT (I know it's the same as above - perhaps update it too?)
```suggestion
    let (method_name, method_name_size) = if let Some(method) = message.method() {
      (method.as_ptr(), method.len())
    } else {
      (std::ptr::null(), 0)
    };
```

ffi/athcon/bindings/rust/athcon-vm/src/lib.rs
```diff
@@ -53,6 +53,7 @@ pub struct ExecutionMessage {
   recipient: Address,
   sender: Address,
   input: Option<Vec<u8>>,
+  method: Option<Vec<u8>>,
```
poszu, line 56: Why not use a String here? 
```suggestion
  method: Option<String>,
```

ffi/athcon/bindings/rust/athcon-vm/src/lib.rs
```diff
@@ -414,6 +435,19 @@ impl TryFrom<&ffi::athcon_message> for ExecutionMessage {
...
+          return Err("msg.method_data is null but msg.method_size is not 0".to_string());
+        }
+        None
+      } else if message.method_name_size == 0 {
+        None
+      } else {
+        Some(from_buf_raw::<u8>(
+          message.method_name,
+          message.method_name_size,
+        ))
```
poszu, line 449: NIT: This and all the uses of `from_buf_raw` could be rewritten without a helper:

```suggestion
        Some(
          unsafe { slice::from_raw_parts(message.method_name, message.method_name_size) }.to_vec(),
        )
```

ffi/athcon/include/athcon/athcon.h
```diff
@@ -112,6 +112,21 @@ extern "C"
      */
     size_t input_size;
 
+    /**
+     * The destination program method to call.
+     *
+     * The arbitrary length byte array of the method name of the call.
+     * This MAY be NULL.
+     */
+    const uint8_t *method_name;
```
poszu, line 121: How about 
```suggestion
    const char *method_name;
```

ffi/athcon/bindings/rust/athcon-vm/src/lib.rs
lrettig: https://github.com/athenavm/athena/issues/113

tests/entrypoint/src/main.rs
```diff
@@ -0,0 +1,36 @@
...
+use athena_vm_sdk::call;
+
+// Note: the test harness installs this contract code at ADDRESS_ALICE
+
+pub fn main() {
+  // recurse forever
+  let address = address_to_32bit_words(ADDRESS_ALICE);
+  let value: [u32; 2] = [0, 0];
+  unsafe { athena_vm::host::call(address.as_ptr(), std::ptr::null(), 0, value.as_ptr()) };
+}
```
poszu, line 16: How about passing the address to call in the input rather than assuming that Host installed this contract at specific address? Something like:
```suggestion
pub fn main() {
  let input = athena_vm::io::read_vec::<Address>();
  let address = bincode::deserialize(&input).unwrap();
  athena_vm_sdk::call(address, Some(input), 0);
}
```

vm/entrypoint/src/syscalls/host.rs
```diff
@@ -3,25 +3,36 @@ use core::arch::asm;
...
+/// the method name in bytes.
 /// `amount` is the number of coins to transfer to the callee.
 /// For now there is no return value and no return status code. The caller can assume
 /// that, if this function returns, the call was successful.
 ///
 /// See https://github.com/athenavm/athena/issues/5 for more information.
 #[allow(unused_variables)]
 #[no_mangle]
-pub extern "C" fn call(address: *const u32, input: *const u32, len: usize, amount: *const u32) {
+pub extern "C" fn call(
```
poszu, line 17: I don't understand why all syscall helpers are marked as `extern "C"`, do you know? They are not called over FFI. They are compiled to risc-v and called normally.

vm/sdk/src/lib.rs
```diff
@@ -1,21 +1,21 @@
...
+  if #[cfg(target_os = "zkvm")] {
+    mod spawn;
+    pub use spawn::spawn;
+    mod call;
+    pub use call::call;
+    mod deploy;
+    pub use deploy::deploy;
+    mod io;
+    pub use io::read;
+    pub use io::write;
```
poszu, line 16:
```suggestion
    pub use io::read_storage;
    pub use io::write_storage;
```

vm/entrypoint/src/syscalls/host.rs
lrettig: We inherited this from sp1, see, e.g.: https://github.com/succinctlabs/sp1/blob/dev/crates/zkvm/entrypoint/src/syscalls/memory.rs

My best guess why this is the case is that some of these syscalls plug directly into the Rust ABI: https://github.com/rust-lang/rust/blob/56e35a5dbb37898433a43133dff0398f46d577b8/library/std/src/sys/pal/zkvm/abi.rs

poszu: It seems that in our case, the syscalls are exported in: https://github.com/athenavm/athena/blob/ae54d14c7c8b489240765cafb393d5266665cdc6/vm/lib/src/lib.rs#L6-L13 to avoid cyclic dependencies between vm/entrypoint and vm/lib crates... The entrypoint depends on lib so lib cannot depend on entrypoint directly, where these syscalls are defined hence it declares that they "exist somewhere" using `export "C"`. IMHO we should really untangle these kind of things as it makes the code complicated and hard to reason about :(

lrettig: #117

tests/entrypoint/src/main.rs
```diff
@@ -25,12 +12,22 @@ pub struct EntrypointTest {};
...
 
   #[callable]
   fn test2() {
+    // no-op
   }
 
+  #[allow(dead_code)]
   fn test3() {
+    // no-op
   }
```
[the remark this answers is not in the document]
lrettig: the point is that this test should fail:

https://github.com/athenavm/athena/pull/112/files#diff-db7734ca10e82c440f78ad4c35a8a418dabdb4afe27289c8aadd86e406899a59R208-R226

In other words, a method not marked `#[callable]` shouldn't be callable

ffi/athcon/bindings/rust/athcon-client/src/host.rs
```diff
@@ -161,7 +162,16 @@ pub unsafe extern "C" fn call(
...
       &msg.sender.bytes,
       &msg.value.bytes,
-      std::slice::from_raw_parts(msg.input_data, msg.input_size),
+      if !msg.input_data.is_null() && msg.input_size > 0 {
+        std::slice::from_raw_parts(msg.input_data, msg.input_size)
+      } else {
+        &[]
+      },
+      if !msg.method_name.is_null() && msg.method_name_size > 0 {
+        std::slice::from_raw_parts(msg.input_data, msg.input_size)
```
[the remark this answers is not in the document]
lrettig: good catch

poszu (approved):

tests/entrypoint/src/main.rs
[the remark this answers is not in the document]
lrettig: correct, but that would be testing something different :) if you think these tests should be restructured/moved, we can do that.

PR #119, opened 2024-10-04
Bugfix in go bindings

lrettig:
@poszu we need better tests for the bindings too...

### DIFF
--- a/ffi/athcon/bindings/go/host.go
+++ b/ffi/athcon/bindings/go/host.go
@@ -84,7 +84,7 @@ type HostContext interface {
 	GetBalance(addr Address) Bytes32
 	GetTxContext() TxContext
 	GetBlockHash(number int64) Bytes32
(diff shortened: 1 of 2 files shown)


poszu:
> @poszu we need better tests for the bindings too...

Would you be willing to add the missing tests together with the fix?

poszu:
TBH, I'm not sure how catch this kind of mistakes. There wasn't any bug per se, the `HostContext` interface just didn't use all the data from the `athcon_message`. Perhaps if `athcon_call_fn` function passed the data as arguments instead of a structure, it would complain about unused parameters, wdyt?

poszu (approved):

lrettig:
> TBH, I'm not sure how catch this kind of mistakes. There wasn't any bug per se, the `HostContext` interface just didn't use all the data from the `athcon_message`. Perhaps if `athcon_call_fn` function passed the data as arguments instead of a structure, it would complain about unused parameters, wdyt?

This is sort of the opposite way of addressing it, by having the host interface work with the same data structures as the underlying C functions: #123. Take a look.

PR #121, opened 2024-10-07
scheduled cargo-audit workflow

poszu:
Add an audit workflow that will run [cargo-audit](https://github.com/rustsec/audit-check):
- 12:00 AM (midnight) UTC every day
- on changes to Cargo.toml and Cargo.lock

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -1,0 +1,17 @@
+name: Security audit
+on:
+  merge_group:
+  schedule:
+    - cron: "0 0 * * *"
+  push:
+    paths:
+      - "**/Cargo.toml"
+      - "**/Cargo.lock"
+jobs:
+  security_audit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: rustsec/audit-check@v2.0.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}


lrettig:
Can we do weekly instead of nightly? Will it keep opening new issues for the same vulnerability if we haven't addressed it yet? Also, curious what this provides on top of dependabot, which is already enabled.

poszu:
> Can we do weekly instead of nightly? Will it keep opening new issues for the same vulnerability if we haven't addressed it yet? 

Sure, but there isn't any benefit of doing it weekly - it doesn't spam in any way, and it won't open new issues for already reported problems. 

>Also, curious what this provides on top of dependabot, which is already enabled.

Dependabot has a different goal. It updates out-of-date dependencies to newer versions. The `cargo-audit` finds vulnerabilities, explains it, and proposes a solution, which can be upgrading or downgrading to a specific version. See https://github.com/spacemeshos/post-rs/issues?q=sort%3Aupdated-desc+is%3Aissue+RUSTSEC for examples of issues it opened.

lrettig (approved):
thanks for the explanations

lrettig:
It looks like we need to commit `Cargo.lock` in order for this to work, see https://github.com/athenavm/athena/actions/runs/11238430868/job/31243174338?pr=125:

image

Historically the advice was to not commit Cargo.lock for libraries, but see https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html

poszu:
@lrettig I've already prepared a PR #124 including the lock files. I think we should include them as we release binaries (the VM shared lib and risc programs) and it's good to know and control what versions of deps we use to build them.

PR #122, opened 2024-10-07
cleanup tests in 'core'

poszu:
Cleanups of tests in the `core` crate. 

Changes:
- removed the wallet test. The same functionality is already tested in its tests located in `examples/wallet/script/src/bin/execute.rs`. The `core` should be unaware of our example wallet implementation. All core VM runtime functionality could be tested with a set of test programs specifically written for this purpose, not re-using examples. Reusing examples causes breaking `core` tests each time the example code is modified
- removed the `host` test as it assumed a specific host implementation, which the runtime shouldn't be aware of. Instead, it uses a mocked interface and asserts specific host calls when testing the syscalls.
	- the host test is now an integration test located in `tests/host/tests/host.rs` (same functionality covered)
- started cleaning up the test-only global constants from `athena-interface` (like `SOME_COINS`, `ADDRESS_CHARLIE` etc.). The idea is that every test should define its own addresses and values to work with.
- removed the default starting state of the `MockHost` to reduce "magic" in tests. The tests are responsible for initializing the host state to their needs themself.


core/src/runtime/mod.rs
```diff
@@ -838,169 +815,22 @@ pub mod tests {
...
   #[test]
-  #[should_panic]
   fn test_panic() {
     let program = panic_program();
     let mut runtime = Runtime::new(program, None, AthenaCoreOpts::default(), None);
-    runtime.execute().unwrap();
-  }
-
-  #[test]
-  fn test_wallet() {
```
lrettig, line 849: Why remove this?

interface/src/lib.rs
```diff
@@ -741,12 +726,14 @@ mod tests {
   #[test]
   fn test_transfer_balance() {
     let mut host = MockHost::new();
+    host.set_balance(&ADDRESS_ALICE, 10000);
     assert_eq!(host.get_balance(&ADDRESS_CHARLIE), 0);
-    assert_eq!(host.get_balance(&ADDRESS_ALICE), SOME_COINS);
+    assert_eq!(host.get_balance(&ADDRESS_ALICE), 10000);
```
lrettig, line 731: Why not use the const value?

core/src/runtime/mod.rs
poszu: Hi, I planned to explain in the description after fixing up the CI. Two reasons:
1. `core` is not the place for programs' tests
2. the same functionality is already tested in the tests in `examples/wallet/script/src/bin/execute.rs` (it's almost identical)

sdk/src/lib.rs
```diff
@@ -124,25 +123,12 @@ mod tests {
     client.execute(elf, stdin, None, None, None).unwrap();
   }
 
-  #[test]
-  fn test_host() {
```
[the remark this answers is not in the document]
poszu: Similar reason, this test is out of place here and it's hard to tell what it is actually testing and why it needs this `host-test` binary for this. It's not even doing any assertions (other than the execution didn't return any error). It's repeating coverage in the `tests/host/tests/host.rs`

interface/src/lib.rs
poszu: Two reasons:
1. IMHO the `athena-interface` is not the right place for these constants (ADDRESS_ALICE, SOME_COINS, STORAGE_KEY, and so on) and the MockHost implementation. They are not part of the VM interface and shouldn't be exposed by this crate - my "far plan" is to get rid of them all.
2. For me, global constants like these in tests only increase the cognitive load by introducing yet another "name" (that needs to be looked up/jumped to definition to a distant place in code/other file or even crate) to understand its value.

I also removed the state of balances and storage initialized by default in the MockHost to reduce "magic" in tests (state coming out of the blue) - instead the tests start with "empty" state and need to initialize the host with whatever things they need. It leads to tests that easier to understand.

core/src/runtime/mod.rs
```diff
@@ -842,169 +819,22 @@ pub mod tests {
...
-      &2103516
-    );
-
-    // now attempt to execute each function in turn
-    // first, the spawn
-    runtime.execute_function("athexp_spawn").unwrap();
-    drop(runtime);
-
-    // get newly-created wallet address
-    let spawn_result = host.get_spawn_result().unwrap().clone();
```
lrettig, line 914: I think we might be able to drop `get_spawn_result` entirely

lrettig (commented):
Let's merge #147, then bring this up to date, then I'll do one final review. LGTM overall.

poszu:
> Let's merge the outstanding PRs (#147, #145, #138), then bring this up to date, then I'll do one final review. LGTM overall.

@lrettig These PRs are now merged, and I updated this PR. Please take a look.

lrettig (approved):
by the way the Go tests here are also passing for me locally

PR #123, opened 2024-10-07
Go bindings data structures

lrettig:
Modifies go host test to use the same underlying data structures as the C types. This is one possible approach to a regression test for the issue fixed in #119. Not sure we actually want to merge this. Another possible approach was described in https://github.com/athenavm/athena/pull/119#issuecomment-2396658412.

This mimics the main host interface on the Rust side:

https://github.com/athenavm/athena/blob/e8b132da055f60acd2ac54fd294834e570ea9ba5/interface/src/lib.rs#L316-L323

although the (unused) Rust bindings are different:

https://github.com/athenavm/athena/blob/e8b132da055f60acd2ac54fd294834e570ea9ba5/ffi/athcon/bindings/rust/athcon-client/src/host.rs#L10-L31

### DIFF
--- a/ffi/athcon/bindings/go/host.go
+++ b/ffi/athcon/bindings/go/host.go
@@ -43,6 +43,73 @@ const (
 	StorageModifiedRestored StorageStatus = C.ATHCON_STORAGE_MODIFIED_RESTORED
 )
 
+type StatusCode int
+
+const (
+	AthconSuccess                   StatusCode = C.ATHCON_SUCCESS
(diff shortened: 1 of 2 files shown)


poszu:
I like that `func call()`  is much cleaner now. However, this change does not resolve the original problem - one can still forget to initialize a new field and Go will not complain about that (as opposed to Rust which wouldn't compile).

lrettig:
assuming for now that we don't want this

PR #125, opened 2024-10-08
clean up logging

poszu:
- remove `setup_logger()` and `setup_tracer()` (which was "deprecated") from `athena-core::utils`. Reasoning: `athena-core` is a library and should not configure logging - this is the responsibility of the application invoking the library (so that it has freedom to choose whatever settings are good).
- add simple `tracing_subscriber` setup in CLI executor and in the unit tests instead,
- changed all usages of the `log` to `tracing` to not mix the two.
- removed `--profile` flag from `CLI execute` as it used a deprecated `athena_core::utils::setup_tracer()` and the same functionality can be obtained by setting `RUST_LOG=trace`
- removed `--verbose` flag from `CLI execute` as it did nothing.

In general, a "utils" package is an anti-pattern and I'd suggest getting rid of the `athena-core::utils` eventually (in steps).


examples/wallet/script/src/bin/execute.rs
```diff
@@ -124,7 +123,7 @@ mod tests {
 
   #[test]
   fn deploy_template() {
-    athena_sdk::utils::setup_logger();
+    env_logger::builer().is_test(true).try_init();
```
lrettig, line 126: `builer` here looks like a typo

lrettig (changes requested):

examples/wallet/script/src/bin/execute.rs
poszu: Yes, thanks! We really should start building and running the /examples and /tests in CI.

poszu:
It needs https://github.com/athenavm/athena-project-template/pull/2 to pass CI

lrettig (approved):
Can we document this somewhere? I know we don't have comprehensive docs yet, but can we at least add something to README or CONTRIBUTING or a new doc outlining how to enable debug/tracing, both on the host and guest sides?

poszu:
> Can we document this somewhere? I know we don't have comprehensive docs yet, but can we at least add something to README or CONTRIBUTING or a new doc outlining how to enable debug/tracing, both on the host and guest sides?

I added a short README in the `/ffi` directory.

PR #137, opened 2024-10-10
remove `athena-hostfunctions` and simplify host syscalls

poszu:


jellonek (approved):

vm/entrypoint/src/syscalls/host.rs
```diff
@@ -41,57 +42,54 @@ pub extern "C" fn call(
...
 /// Read from host storage at the given address and key.
-///
-/// The output is stored in the `key` pointer.
 #[allow(unused_variables)]
-#[no_mangle]
-pub extern "C" fn read_storage(key: *mut u32) {
+pub fn read_storage(key: &[u32; 8]) -> [u32; 8] {
   #[cfg(target_os = "zkvm")]
   unsafe {
+    let mut result = *key;
```
lrettig, line 49: Just to be clear, this does overwrite the value at `key` correct? Wouldn't it be safer and cleaner here to allocate a new memory location instead? As you do in `get_balance`

vm/entrypoint/src/syscalls/host.rs
```diff
@@ -41,57 +42,54 @@ pub extern "C" fn call(
...
   unsafe {
+    let mut balance = std::mem::MaybeUninit::<u64>::uninit();
     asm!(
         "ecall",
         in("t0") crate::syscalls::HOST_GETBALANCE,
-        in("a0") value,
-    )
+        in("a0") balance.as_mut_ptr(),
+    );
+    balance.assume_init()
```
lrettig, line 92: possibly dumb question: how do we ensure that the value is read here as LE? I know the value is written that way, but what if the host architecture here is BE?

vm/entrypoint/src/syscalls/host.rs
```diff
@@ -41,57 +42,54 @@ pub extern "C" fn call(
...
+    let status: u32;
     asm!(
         "ecall",
         in("t0") crate::syscalls::HOST_WRITE,
-        in("a0") key,
-        in("a1") value,
-    )
+        in("a0") key.as_ptr(),
+        in("a1") value.as_ptr(),
+        lateout("t0") status,
```
lrettig, line 73: same question as in `get_balance`, how do we explicitly ensure we're reading LE here?

tests/host/src/main.rs
```diff
@@ -2,81 +2,67 @@
...
 
 pub fn main() {
-  let mut key = bytes32_to_32bit_words(STORAGE_KEY);
+  let key = bytes32_to_32bit_words(STORAGE_KEY);
   let value = bytes32_to_32bit_words(STORAGE_VALUE);
   let value2: [u32; 8] = [0xaa; 8];
-  let value_unset: [u32; 8] = [0; 8];
   let address_charlie = address_to_32bit_words(ADDRESS_CHARLIE);
 
   // note: for all of these calls, the result is written to the first argument, hence as_mut_ptr()
```
lrettig, line 20: remove this comment, no longer accurate

core/src/syscall/host.rs
```diff
@@ -50,11 +50,7 @@ impl Syscall for SyscallHostWrite {
       &Bytes32Wrapper::from(value).into(),
     );
 
-    // save return code
-    let mut status_word = [0u32; 8];
-    status_word[0] = status_code as u32;
-    ctx.mw_slice(arg1, &status_word);
-    Ok(Outcome::Result(None))
+    Ok(Outcome::Result(Some(status_code as u32)))
```
lrettig, line 53: same question again, do we need to ensure this is written as LE explicitly?

lrettig (changes requested):

vm/entrypoint/src/syscalls/host.rs
poszu: It doesn't overwrite `key` in any place now. It creates a mutable copy of the key on the stack named `result`, which the pointer to is passed in the register (same as before). The only difference is that the code now creates a copy instead of overwriting the key, which is returned from the function "normally".

vm/entrypoint/src/syscalls/host.rs
poszu: We don't and I think can't ensure this. However, I suppose none of the program <-> VM communication would work if the host machine was BE. For example, if the program wrote a `0x000000CA` into a register, the host would read `0xCA000000` and nothing would work. To support a BE host machine, I think we would need to always go through a LE (program) <-> native (host) conversion (sometimes a noop, if host is also LE), whenever we read and write registers and program memory on the VM side.

core/src/syscall/host.rs
poszu: No (and we also don't do this in many other places).

vm/entrypoint/src/syscalls/host.rs
lrettig: We can always ensure this if we want to but it's extra work. See e.g. the conversion code here, using `bytemuck`: https://github.com/risc0/risc0/blob/0e6169b2fe652cc3e62927a5347b4af496ad62dd/risc0/zkvm/src/host/server/testutils.rs#L22-L37.

We're not alone in this issue: https://github.com/risc0/risc0/issues/420. I think it's fine for now. I opened #141 so it's documented/searchable. It would be good to point out places in the code that make this assumption, unless you think there are just too many of them.

lrettig (approved):

vm/entrypoint/src/syscalls/host.rs
poszu: I think it's every place where the VM reads or writes a register or program memory.

PR #138, opened 2024-10-10
remove redundant stack-depth test

poszu:
The same test functionality can be achieved using the recursive-call test.


jellonek (approved):

runner/src/vm.rs
```diff
@@ -336,17 +336,18 @@ mod tests {
...
 
     let client = ExecutionClient::new();
-    let elf = include_bytes!("../../tests/stack_depth/elf/stack-depth-test");
-    let stdin = AthenaStdin::new();
-    let vm = AthenaVm::new();
-    let mut host = MockHost::new_with_vm(&vm);
-    host.deploy_code(ADDRESS_ALICE, elf.to_vec());
+    let elf = include_bytes!("../../tests/recursive_call/elf/recursive-call-test");
+    let mut stdin = AthenaStdin::new();
+    stdin.write::<u32>(&7);
```
lrettig, line 341: It's not obvious or intuitive why this causes a `CallDepthExceeded` error. `MockHost` has a hardcoded max stack depth of 10. Why does running fib(7) cause this?

poszu: It uses a mocked `HostInterface` that returns the expected error from `call()`:

```rust
host
      .expect_call()
      .returning(|_| ExecutionResult::new(StatusCode::CallDepthExceeded, 0, None, None));
```

lrettig: Hmm. I see. But this isn't really testing recursion then, is it? Like, we _also_ want to test the code path where the host accurately keeps track of the stack depth and actually triggers the error when it's exceeded.

You can do that with fib() of course but it will be much slower than the stack depth test you removed.

poszu: Hmm, no it isn't. I could make it pass a few calls before it returns `CallDepthExceeded`. However, in my opinion, this is not a place for a test that checks counting depth - this happens currently in the CALL syscall in `core`.

lrettig: Not sure what you mean. The checking happens in the host, not in the syscall:

https://github.com/athenavm/athena/blob/a9cc278cc47d96f8e08eac1806410a24a4b14d20/core/src/syscall/host.rs#L84

And I don't see a test in `core/src/syscall/host.rs` for this. Can we add one in tandem to this change?

poszu: I mean that keeping track of the calls depth takes place in the CALL syscall here: https://github.com/athenavm/athena/blob/a9cc278cc47d96f8e08eac1806410a24a4b14d20/core/src/syscall/host.rs#L143 The host is responsible for checking if the call stack is not too deep. If the test intends to check if the `MockHost` implementation returns `CallDepthExceeded` status code, it's not the right place to do this - this kind of test should be placed (if at all - it is a mock after all :shrug:) next to the `MockHost` implementation.

>And I don't see a test in core/src/syscall/host.rs for this. Can we add one in tandem to this change?

Yeah, it's worth having a test that verifies this in the `core` crate.

lrettig: Fair point, it is the host's responsibility to check, not the VM. But I think we can still check two things on the VM side:

- assuming the host reports that the call depth is exceeded, make sure the error is handled/bubbled up correctly in the VM
- in a recursive call, make sure the msg.depth goes up by one

poszu: Added tests - @lrettig please take another look :pray:

lrettig (approved):

PR #139, opened 2024-10-10
Use SCALE instead of Borsh

### LINKED ISSUE #136: RUSTSEC-2024-0370: proc-macro-error is unmaintained
> proc-macro-error is unmaintained

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Status              | unmaintained                |
| Package             | `proc-macro-error`                      |
| Version             | `1.0.4`                   |
| URL                 | [https://gitlab.com/CreepySkeleton/proc-macro-error/-/issues/20](https://gitlab.com/CreepySkeleton/proc-macro-error/-/issues/20) |
| Date                | 2024-09-01                         |

proc-macro-error&#39;s maintainer seems to be unreachable, with no commits for 2 years, no releases pushed for 4 years, and no activity on the GitLab repo or response to email.

proc-macro-error also depends on `syn 1.x`, which may be bringing duplicate dependencies into dependant build trees.

## Possible Alternative(s)

- [manyhow](https://crates.io/crates/manyhow)
- [proc-macro-error2](https://crates.io/crates/proc-macro-error2)
- [proc-macro2-diagnostics](https://github.com/SergioBenitez/proc-macro2-diagnostics)

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2024-0370.html) for additional details.

poszu:
Switch to scale codec for encoding & decoding to be in line with spacemesh, which uses scale extensively for serialization and doesn't use borsh. 

Also, fixes #136 (borsh depends on an unmaintained crate reported in the sec issue).


core/src/runtime/mod.rs
```diff
@@ -935,7 +935,7 @@ pub mod tests {
     let ctx = AthenaContext::new(wallet_template, ADDRESS_ALICE, 0);
     let mut runtime = Runtime::new(program.clone(), Some(&mut host), opts, Some(ctx));
     stdin.write_vec(spawn_result.blob.clone());
-    stdin.write_vec(to_vec(&send_args).expect("failed to serialize wallet send arguments"));
+    stdin.write_vec(send_args.encode());
```
lrettig, line 938: out of curiosity, how does error handling happen here? does `encode()` panic on error?

lrettig (approved):

core/src/runtime/mod.rs
poszu: As far as I understand, encoding cannot fail (except for running out of memory when the allocator fails, when growing the Vec and Rust will panic in this scenario "by itself").

PR #142, opened 2024-10-10
Add maxspend API

### LINKED ISSUE #128: API: MaxSpend
Given a transaction, go-spacemesh needs to know the maximum amount that it can spend from the principal account for mempool management purposes. More specifically, given context including `target_template` and an Athena payload containing

`METHOD | ARGS{RECIPIENT (Address) | AMOUNT (uint64)}`

this method needs to return a uint64 value. The template should be responsible for calculating this based on the tx payload. Accordingly, the single sig wallet template should implement a `max_spend()` method with very straightforward logic akin to the existing go-spacemesh genvm logic: https://github.com/spacemeshos/go-spacemesh/blob/118ee7f539fcc0a8b20f2cea672091aab78eb7ce/genvm/templates/wallet/wallet.go#L24-L34.

lrettig:
Adds a `maxspend` function to the VM, available over FFI, to fix #128


lrettig:
Closing in favor of #158

PR #144, opened 2024-10-11
Update C API to use uint64_t instead of uint256be where it makes sense

### LINKED ISSUE #140: Return u64 from `get_balance` in the athcon FFI
Currently, the `get_balance()` function returns a `athcon_uint256be`:
https://github.com/athenavm/athena/blob/96b9afa3f8e08e24506b5466218842354eb40afb/ffi/athcon/include/athcon/athcon.h#L549-L559

Which is supposed to be a 256bit big-endian integer encoded in a byte array:
https://github.com/athenavm/athena/blob/96b9afa3f8e08e24506b5466218842354eb40afb/ffi/athcon/include/athcon/athcon.h#L43-L46

There are two problems with it:
- the athena balance is `u64` so its inefficient to pass it as 256bits
- it's supposed to be big-endian (according to the type description) but we put the value as little-endian there

IMHO, we should change the C API to return a uint64_t instead to avoid confusion, make it efficient and simplify the code.

poszu:
Closes #140.
Updated:
 - `get_balance()` to return `uint64_t`
 - `athcon_message::value` to `uint64_t` as it represents coins and is converted to uint64_t internally anyways
 - `athcon_tx_context::tx_gas_price` as it is converted to `uint64_t` internally

### DIFF
--- a/ffi/athcon/bindings/go/athcon.go
+++ b/ffi/athcon/bindings/go/athcon.go
@@ -171,7 +171,7 @@ func (vm *VM) Execute(
 	recipient, sender Address,
 	input []byte,
 	method []byte,
-	value Bytes32,
+	value uint64,
 	code []byte,
 ) (res Result, err error) {
 	if len(code) == 0 {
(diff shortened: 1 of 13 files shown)


lrettig (commented):
Why not also update athcon.h and the Rust side?

poszu:
> Why not also update athcon.h and the Rust side?

What do you mean? I updated athcon.h and both Rust sides of the FFI in this PR (and Go host side too).

lrettig (approved):

lrettig:
> > Why not also update athcon.h and the Rust side?
> 
> What do you mean? I updated athcon.h and both Rust sides of the FFI in this PR (and Go host side too).

I must've only been seeing one commit, oops...

PR #145, opened 2024-10-11
add athcon_encode_tx to serialize transactions

### LINKED ISSUE #131: API: Encode tx payload [spawn, spend]
For tests, go-spacemesh needs to be able to generate valid Athena transactions. More specifically, Athena needs to expose a function that, given a principal account address, a principal template, a nonce, a method selector, and method/template-specific args, returns an encoded Athena payload that can be "loaded" into a tx and signed.

For the single sig wallet, for now, we need this to work for two methods, spawn and spend:

spawn: owner pubkey (bytes32)
spend: recipient (Address), amount (uint64)

poszu:
Fixes #131.

Adds 3 functions to the C API at "library level" - they don't require having a VM or host instance:
- `athcon_encode_tx_spawn`: encodes `spawn` payload (method selector + arguments (pubkey))
- `athcon_encode_tx_send`: encodes `spend` payload (method selector + recipient address and amount)
- `athcon_free_bytes` to free the heap-allocated bytes returned from the encoding functions

Renamed `WalletProgram::Send` to `Spend` as well as its arguments.

Added the 4B `MethodSelector` and TX `Payload` (method + arguments) duplicating some of the effort from #147  because it was needed here. Will need to resolve conflicts here or there depending which is merged first.


interface/src/transaction.rs
```diff
@@ -0,0 +1,34 @@
+pub use parity_scale_codec::{Decode, Encode};
+
+use crate::Address;
+
+#[derive(Debug, Clone, Encode, Decode)]
+pub struct Transaction {
```
lrettig, line 6: I wouldn't call this `Transaction`, I'd call it `TransactionPayload` or simply `Payload`. It's the inner part of the tx that's opaque to the host. It consists of just method and args.

- nonce will be handled by the host for now; we'll move it into the VM later
- principal needs to be readable by the host, so it sits outside this payload. It's the `sender` in the execution message.
- template isn't needed here; the template `code` is also part of the execution message.

ffi/athcon/include/athcon/athcon.h
```diff
@@ -820,6 +820,27 @@ extern "C"
...
 
+  typedef struct athcon_bytes_t
+  {
+    const uint8_t* ptr;
+    size_t size;
+  } athcon_bytes;
+
+  void athcon_free_bytes(athcon_bytes* v);
+
+  athcon_bytes* athcon_encode_tx(
```
lrettig, line 831:
```suggestion
  athcon_bytes* athcon_encode_payload(
```

ffi/athcon/include/athcon/athcon.h
```diff
@@ -820,6 +820,27 @@ extern "C"
...
+    const athcon_address* principal_account,
```
lrettig, line 832: As mentioned above, this only needs to take method, method_size, args, and args_size.

ffi/athcon/bindings/rust/athcon-vm/src/encode_tx.rs
```diff
@@ -0,0 +1,137 @@
...
+/// - `template` - Optional transaction's template account.
+/// - (`method`, `method_size`) - Optional method to call
+/// - `nonce` - The transaction's nonce
+/// - (`args`, `args_size`) - Optional transaction's serialized arguments.
+///
+/// # Safety
+/// - the caller is responsible for freeing the returned vector
+///   via the `athcon_free_vector` function.
+#[no_mangle]
+unsafe extern "C" fn athcon_encode_tx(
```
lrettig, line 19: I think we can do away with this method entirely and just keep the two encode txs below, one for spawn, one for spend. They need to include method as well as args.

ffi/athcon/bindings/rust/athcon-vm/src/encode_tx.rs
```diff
@@ -0,0 +1,137 @@
...
+  pubkey: *const ffi::athcon_bytes32,
+) -> *mut ffi::athcon_bytes {
+  let encoded = Pubkey((*pubkey).bytes).encode();
+  let (ptr, size) = crate::allocate_output_data(encoded);
+
+  Box::into_raw(Box::new(ffi::athcon_bytes { ptr, size }))
+}
+
+#[no_mangle]
+unsafe extern "C" fn athcon_encode_tx_send(
```
lrettig, line 59:
```suggestion
unsafe extern "C" fn athcon_encode_tx_spend(
```

ffi/athcon/bindings/rust/athcon-vm/src/encode_tx.rs
```diff
@@ -0,0 +1,137 @@
...
+  recipient: *const ffi::athcon_address,
+  amount: u64,
+) -> *mut ffi::athcon_bytes {
+  let args = SendArguments {
```
lrettig, line 63: should probably rename this struct `SpendArguments` for consistency with go-spacemesh:

https://github.com/spacemeshos/go-spacemesh/blob/5a40816d2bd503f53c9fa5a0476fe7fd2d8506db/genvm/templates/wallet/types.go#L14-L18

lrettig:
Awesome! Proposing a few tweaks here, LMK what you think:

> * `athcon_encode_tx`: serializes transaction into heap-allocated raw bytes given:
>   
>   * principal account address
>   * optional template address
>   * optional method (bytes)
>   * optional program arguments (bytes)
>   * nonce
> * `athcon_encode_tx_spawn`: encodes spawn arguments (pubkey)
> * `athcon_encode_tx_send`: encodes `SendArguments` (recipient address and amount)
> * `athcon_free_bytes` to free the heap-allocated bytes returned from `athcon_encode_tx`

As indicated in the review comments, I don't think we need athena to encode an entire tx. It just needs to encode a _payload_ that contains the opaque portion that the host doesn't need to know about -- for now, this is just method selector and args. See https://github.com/athenavm/athena/pull/145#discussion_r1797297694 for rationale.

So I think we can drop `athcon_encode_tx` entirely and modify `athcon_encode_tx_{spend|spawn}` to encode both method and args.

> * Is the approach to make it a "free standing" function in the VM library a good choice? One only needs to load the library to call it, there is no need to create a VM instance.

Sounds good - these methods definitely do not require executing VM code. (Note that the situation will be different for `validate()` and `maxspend()`.)

> * How does the caller obtain the serialized program arguments (the `args` argument) for all other programs in general? The VM library cannot serialize these as it cannot know all programs.

It's fine to have explicit encoding methods for these "special" tx types, but in future we obviously can't have explicit encoding methods for every possible combination of method and args, once we have many templates, user-deployed templates, etc. At that stage we'll need a separate SDK that, given program metadata (akin to EVM's contract ABI) and input args, returns an encoded payload.

ffi/athcon/include/athcon/athcon.h
```diff
@@ -820,6 +820,27 @@ extern "C"
...
+    athcon_address* template_addresss,
+    uint8_t* method,
+    size_t method_size,
+    uint64_t nonce,
+    uint8_t* args,
+    size_t args_size
+  );
+
+  athcon_bytes* athcon_encode_tx_spawn(const athcon_bytes32* pubkey);
```
lrettig, line 841: What do we do about error handling here? Does the caller just assume that an empty return value indicates error?

poszu: Scale encoding is infallible.

ffi/athcon/include/athcon/athcon.h
poszu: Thinking ahead, considering we go with the 4B method selector. Would you rather have this function take the selector or method name as a string?

ffi/athcon/bindings/rust/athcon-vm/src/encode_tx.rs
poszu: I used "send" because the method is called send. Would you rather rename the method to spend?

poszu: Change to "spend" everywhere.

ffi/athcon/bindings/go/athcon.go
```diff
@@ -64,18 +62,19 @@ const (
...
+type Library struct {
 	// handle to the opened shared library. Must be closed with Dlclose.
 	libHandle uintptr
-	// handle to the VM instance. Must be destroyed with athcon_destroy.
-	handle *C.struct_athcon_vm
+
+	create func() *C.struct_athcon_vm
+
+	encodeTxSpawn func(*C.athcon_bytes32) *C.athcon_bytes
+	encodeTxSend  func(*C.athcon_address, C.uint64_t) *C.athcon_bytes
```
lrettig, line 72:
```suggestion
	encodeTxSpend  func(*C.athcon_address, C.uint64_t) *C.athcon_bytes
```

ffi/athcon/bindings/go/athcon.go
```diff
@@ -85,14 +84,40 @@ func Load(path string) (*VM, error) {
...
 
-	var athcon_create func() *C.struct_athcon_vm
-	purego.RegisterLibFunc(&athcon_create, libHandle, "athcon_create_"+vmName)
-	vmHandle := athcon_create()
+	lib := &Library{
+		libHandle: libHandle,
+	}
+	purego.RegisterLibFunc(&lib.create, libHandle, "athcon_create_"+vmName)
+	purego.RegisterLibFunc(&lib.encodeTxSpawn, libHandle, "athcon_encode_tx_spawn")
+	purego.RegisterLibFunc(&lib.encodeTxSend, libHandle, "athcon_encode_tx_spend")
```
lrettig, line 92:
```suggestion
	purego.RegisterLibFunc(&lib.encodeTxSpend, libHandle, "athcon_encode_tx_spend")
```

ffi/athcon/bindings/go/athcon.go
```diff
@@ -230,18 +255,32 @@ func (vm *VM) Execute(
...
+}
+
+func (l *Library) EncodeTxSpawn(pubkey Bytes32) []byte {
+	encoded := l.encodeTxSpawn(athconBytes32(pubkey))
+	defer l.freeBytes(encoded)
+	tx := C.GoBytes(unsafe.Pointer(encoded.ptr), C.int(encoded.size))
+	return tx
+}
+
+func (l *Library) EncodeTxSend(principal Address, nonce uint64) []byte {
```
lrettig, line 281:
```suggestion
func (l *Library) EncodeTxSpend(principal Address, nonce uint64) []byte {
```

ffi/athcon/bindings/go/athcon.go
```diff
@@ -230,18 +255,32 @@ func (vm *VM) Execute(
...
+	encoded := l.encodeTxSend(athconAddress(principal), C.uint64_t(nonce))
```
lrettig, line 282:
```suggestion
	encoded := l.encodeTxSpend(athconAddress(principal), C.uint64_t(nonce))
```

ffi/athcon/bindings/go/athcon_test.go
```diff
@@ -86,3 +86,22 @@ func TestErrorMessage(t *testing.T) {
...
+	require.NoError(t, err)
+	t.Run("spawn", func(t *testing.T) {
+		tx := lib.EncodeTxSpawn(Bytes32{9, 8, 7, 6})
+		require.NotEmpty(t, tx)
+
+		tx2 := lib.EncodeTxSpawn(Bytes32{1, 2, 3, 4})
+		require.NotEqual(t, tx, tx2)
+	})
+	t.Run("spend", func(t *testing.T) {
+		tx := lib.EncodeTxSend(Address{1, 2, 3, 4}, 191239)
```
lrettig, line 101:
```suggestion
		tx := lib.EncodeTxSpend(Address{1, 2, 3, 4}, 191239)
```

ffi/athcon/bindings/go/athcon_test.go
```diff
@@ -86,3 +86,22 @@ func TestErrorMessage(t *testing.T) {
...
+		require.NotEmpty(t, tx)
+
+		tx2 := lib.EncodeTxSend(Address{1, 2, 3, 4}, 80972)
```
lrettig, line 104:
```suggestion
		tx2 := lib.EncodeTxSpend(Address{1, 2, 3, 4}, 80972)
```

interface/src/transaction.rs
```diff
@@ -0,0 +1,21 @@
+pub use parity_scale_codec::{Decode, Encode};
```
lrettig, line 1: nit: I'd rather this file be called `payload.rs` since the term transaction is a bit confusing here

lrettig (changes requested):

ffi/athcon/include/athcon/athcon.h
lrettig: I guess we can just combine args and method selector into a single byte array as in #147?

poszu: removed this function completely and left encode_spend and encode_spawn

vm/sdk/src/wallet.rs
```diff
@@ -0,0 +1,61 @@
...
+
+#[cfg(test)]
+mod tests {
+  use parity_scale_codec::{Decode, Encode, IoReader};
+
+  use super::SpendArguments;
+
+  #[test]
+  fn encode_decode_spend() {
+    let wallet = Vec::<i32>::from([1, 2, 3, 4]);
```
lrettig, line 45: I'm confused, why is this a valid wallet state? Why not actually instantiate the wallet here and encode the instance?

vm/sdk/src/wallet.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+pub trait WalletProgram {
+  fn spawn(owner: Pubkey) -> Address;
+  fn spend(&self, args: SpendArguments);
+  fn proxy(&self, destination: Address, args: &[u8]);
+  fn deploy(&self, code: Vec<u8>) -> Address;
+}
+
+pub fn encode_spend(state: Vec<u8>, args: SpendArguments) -> Vec<u8> {
```
lrettig, line 21: Why does encoding require the state? I thought encoding consisted of method + args (i.e., the payload type we defined)

lrettig: I see that the method selector is being applied at the higher level, okay, but I'm still confused why this requires state.

poszu: Because the encoded state of the program is also one of the inputs and must be written to the "stdin" along with the actual arguments so that the program side can read it from the IO and decode. For example, look at the spend method definition:
```rust 
fn spend(&self, args: SpendArguments)
```

Both `self` and the `args` come from the "message input".

vm/sdk/src/wallet.rs
poszu: It's a dummy state for testing - it doesn't need to be valid for the sake of this test.

> Why not actually instantiate the wallet here and encode the instance?

For simplicity and to avoid cyclic dependencies. The `struct Wallet` we have in examples/wallet/program, which implements `WalletProgram`, is only one of the possible implementations (and it already imports this pkg so trying to import it here would cause a cyclic dep).

vm/sdk/src/wallet.rs
lrettig: Okay, but it's important to distinguish between the state and the payload, which consists of the method selector and args only. Maybe we need yet another layer here to encapsulate all of these things? It's important because the tx itself contains the payload, but it does not contain the state (which could be quite large). The host (go-sm) has to pass the state into the VM.

By the way, it turns out that spend() actually does not depend on the state:

https://github.com/athenavm/athena/blob/76e6130e26527a61cbf86b8a5950ed91273a69a5/examples/wallet/program/src/main.rs#L41-L46

so it could be turned into an "associated function" (one that doesn't take self) like spawn. It depends on _context_ but not on _state._ verify(), on the other hand, does depend on state.

vm/sdk/src/wallet.rs
lrettig: I just find this test kinda counterintuitive and hard to read, e.g.:

```
let decoded_wallet = Vec::<i32>::decode(&mut input_reader).unwrap();
```

This isn't a "decoded wallet" at all, it's a vector of arbitrary integers.

I think it could be made more readable with some explanatory comments.

lrettig:
This fixes #131 right? Anything still missing?

vm/sdk/src/wallet.rs
poszu: >Okay, but it's important to distinguish between the state and the payload, which consists of the method selector and args only. Maybe we need yet another layer here to encapsulate all of these things? It's important because the tx itself contains the payload, but it does not contain the state (which could be quite large). The host (go-sm) has to pass the state into the VM.

Right, the TX payload contains `method_selector | arguments` but cannot contain `state` so encoding it `athcon_encode_spend()` makes no sense. On the other hand, the `runner`  must pass the state to stdin in its `execute()` method:

```rust
let mut stdin = AthenaStdin::new();
// same as stdin.write_vec(state | args);
stdin.write_vec(state);
stdin.write_vec(args);
```

So the question is, how does the `runner` obtain the state? Should it call the `host` or assume that `msg.input_data` already contains it (i.e., `input_data = state | execution_payload(method | args)`)?

>By the way, it turns out that spend() actually does not depend on the state:

Then, what is the purpose of `spend()` beyond proxying to the `host::call()`? What logic could a contract implement there? I'm trying to understand why this method is even needed in the `WalletProgram`.

vm/sdk/src/wallet.rs
poszu: > This isn't a "decoded wallet" at all, it's a vector of arbitrary integers.

A wallet structure could be anything :shrug:. This:
```rust
struct Foo(Vec<i32>);
impl WalletProgram for Foo {
...
}
```

is a valid wallet, which state consists of a vector of i32s. 

I can change it to a more "natural"  `struct Wallet {...}` if this is confusing.

poszu: Changed to a Wallet struct for clarity.

vm/sdk/src/wallet.rs
lrettig: > Then, what is the purpose of spend() beyond proxying to the host::call()? What logic could a contract implement there? I'm trying to understand why this method is even needed in the WalletProgram.

There's more complex logic for multisig: there can be spend limits per pubkey, daily spend limits, etc.

Cargo.toml
```diff
@@ -18,7 +18,6 @@ members = [
   "vm/sdk",
 ]
 exclude = ["examples"]
-resolver = "2"
```
lrettig, line 21: Was this intentional? I'm getting warnings:

```
warning: virtual workspace defaulting to `resolver = "1"` despite one or more workspace members being on edition 2021 which implies `resolver = "2"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
note: for more details see https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions
```

vm/sdk/src/wallet.rs
lrettig: Let's discuss the rest at https://github.com/athenavm/athena/pull/157#discussion_r1804309118

Cargo.toml
poszu: I don't understand how it works. I thought that "2" was the default, but apparently, it is not. I reverted, and I will come back to it later.

lrettig (approved):

interface/src/payload.rs
```diff
@@ -0,0 +1,82 @@
...
+
+use crate::MethodSelector;
+
+/// Execution payload is passed as input to VM to execute a metod.
+/// The optional state contains a serialized wallet state. It should be
+/// None if the call doesn't require any state.
+#[derive(Clone, Debug, Decode, Default, Encode, PartialEq, Eq)]
+pub struct ExecutionPayload {
+  pub state: Vec<u8>,
+  pub payload: Vec<u8>,
```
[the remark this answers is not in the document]
poszu: I wasn't sure about this TBH. There is a tradeoff here: from the VM perspective it makes sense to put `Payload` type here, it simplifies the VM runner code. On the other hand, the host side always gets `payload` already **encoded** (encoded payload is part of a transaction). So, if `payload` was of type `Payload` instead of a `Vec<u8>`, the host side would have two options:

1. manually scale-encode Execution payload by encoding the raw state and appending the already encoded `payload` it got from TX. (A scale encoding of a struct is simply a concatenation of encodings of all its fields)
2. decode payload, put it into an `ExecutionPayload` struct, and encode the whole thing again.

In both cases the result is equivalent.

I optimized the code for host convenience here, but now that I think of it, it's probably wrong because it led to this weird `ExecutionPayload` with an opaque payload. I will change it to:
```rust
pub struct ExecutionPayload {
  pub state: Vec<u8>,
  pub payload: Payload,
}
```

and update the host code. Let's see which version is better.

interface/src/lib.rs
```diff
@@ -613,6 +609,23 @@ impl HostInterface for MockHost<'_> {
       // create an owned copy of VM before taking the host from self
       let vm = self.vm;
 
+      // The optional msg.input_data  must be enriched with optional account state
+      // and then passed to the VM.
+      let msg = match msg.input_data {
```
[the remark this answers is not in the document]
poszu: It could, but I find a `match` easier to read than `if let Some(data) = msg.input_data {...} else {...};`

lrettig (approved):

interface/src/payload.rs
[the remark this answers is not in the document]
lrettig: I think the change is fine. Also - we don't have to use the same exact data structures on both host and VM side!

interface/src/payload.rs
```diff
@@ -17,6 +17,27 @@ impl From<ExecutionPayload> for Vec<u8> {
   }
 }
 
+impl ExecutionPayload {
+  /// Manually encode using an already encoded payload.
+  /// Effectively: encode(state) | payload
+  pub fn encode_with_encoded_payload<S, P>(state: S, payload: P) -> Vec<u8>
```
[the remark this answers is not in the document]
poszu: Because it's logically connected, it encodes `ExecutionPayload` using half-backed parts.

PR #147, opened 2024-10-11
Consolidate method and args into a single opaque payload

### LINKED ISSUE #113: Use fixed-length method selector
In #112 we added message name to the execution message. Rather than using variable-length strings, let's move to a fixed-length selector. EVM uses four bytes, see: https://solidity-by-example.org/function-selector/.

For simplicity we can use something like a four byte hash of the method name.

### LINKED ISSUE #146: Merge execution message `input` and `method` into single `payload`
As discussed in https://github.com/athenavm/athena/pull/145#issuecomment-2407943175, we need the execution message passed into athena from the host to contain a _single_ opaque `payload` field, rather than distinct `method` and `input` fields. This payload consists of the encoded method selector (#113) prepended to the encoded args. In this fashion we ensure that the host never needs to be aware of the method or args, doesn't need to decode them, etc.

Note that the one possible exception here is a spawn transaction, which requires special handling. Since we don't currently allow a principal to sponsor a call to a method on a different template -- which would allow principal X to pay to spawn principal Y -- effectively the only form of spawn that we support is self-spawn, whereby a principal pays to spawn itself.

In the case of a self-spawn tx, the host knows the tx is a self-spawn because its principal isn't spawned yet. In this case, it independently calculates the principal based on the args (owner pubkey) to make sure it matches the tx principal. (And it can safely assume that it's a self-spawn of a single sig wallet, since that's the only template we support for now.)

For now, we're leaving principal computation inside the host (#130). In the case of a self-spawn tx, we can just pass the entire payload into `ComputePrincipal`. This effectively means updating the `ComputePrincipal` algorithm on the host so that instead of

`BLAKE3(SCALE(template_address | args))[len(output)-20:]`

it does

`BLAKE3(method_spawn | args(pubkey))`

which seems fine for now.

In future when we have more templates and allow proxied calls/spawns, we'll need to modify self-spawn txs so that, as in genvm, they explicitly include a template address. And we'll need to modify `ComputePrincipal` so that it includes the template address, principal address, and principal nonce, in addition to the args.

lrettig:
Fixes #146, also fixes #113


interface/src/lib.rs
```diff
@@ -5,16 +5,67 @@
...
 pub const BYTES32_LENGTH: usize = 32;
+pub const METHOD_SELECTOR_LENGTH: usize = 4;
 pub type Address = [u8; ADDRESS_LENGTH];
 pub type Balance = u64;
 pub type Bytes32 = [u8; BYTES32_LENGTH];
 pub type Bytes = [u8];
+pub type MethodSelectorBytes = [u8; METHOD_SELECTOR_LENGTH];
+pub const METHOD_SELECTOR_DEFAULT: MethodSelector = MethodSelector {
+  bytes: [0u8; METHOD_SELECTOR_LENGTH],
+};
```
poszu, line 23: It wouldn't be needed with the proposed changes to use `Option<MethodSelector>` to indicate "no method".
```suggestion
```

interface/src/lib.rs
```diff
@@ -5,16 +5,67 @@
...
+
+#[derive(Debug, Clone, Deserialize, Eq, PartialEq, PartialOrd, Serialize)]
+pub struct MethodSelector {
+  bytes: MethodSelectorBytes,
+}
```
poszu, line 28: A newtype pattern fits here perfectly:
```suggestion
#[derive(Debug, Clone, Deserialize, Eq, PartialEq, PartialOrd, Ord, Serialize)]
pub struct MethodSelector([u8; METHOD_SELECTOR_LENGTH]);
```

Note: I propose to use `[u8; METHOD_SELECTOR_LENGTH]` instead of `MethodSelectorBytes` because the latter obscures what is going on and abstraction is not needed here as this is an internal detail.

interface/src/lib.rs
```diff
@@ -5,16 +5,67 @@
...
+
+impl MethodSelector {
+  pub fn bytes(&self) -> MethodSelectorBytes {
+    self.bytes
+  }
+}
```
poszu, line 34: AFAICT, this is only used to log the selector in hex. How about instead implementing `std::fmt::Display`?
```suggestion
impl std::fmt::Display for MethodSelector {
  fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
    write!(f, "{}", hex::encode(self.0))
  }
}
```

interface/src/lib.rs
```diff
@@ -5,16 +5,67 @@
...
+impl From<&str> for MethodSelector {
+  fn from(value: &str) -> Self {
+    let h = hash(value.as_bytes());
+    let vbytes = h.as_bytes();
+    MethodSelector {
+      bytes: vbytes[vbytes.len() - METHOD_SELECTOR_LENGTH..]
+        .try_into()
+        .unwrap(),
+    }
+  }
```
poszu, line 45: NIT: it would be simpler to take the first bytes, wdyt?
```suggestion
    MethodSelector(..METHOD_SELECTOR_LENGTH].try_into().unwrap())
  }
```

interface/src/lib.rs
```diff
@@ -5,16 +5,67 @@
...
+      bytes: value.try_into().unwrap(),
+    }
+  }
+}
+
+impl From<MethodSelector> for MethodSelectorBytes {
+  fn from(value: MethodSelector) -> MethodSelectorBytes {
+    value.bytes
+  }
+}
```
poszu, line 60: I think these two aren't needed.
```suggestion
```

poszu (changes requested):

interface/src/lib.rs
lrettig: yes, you're right that it's simpler and it makes no difference

vm/sdk/src/call.rs
```diff
@@ -1,37 +1,30 @@
...
   // we can pad the input but need to know more about the contents
   let input32 = if let Some(input) = input {
     assert!(input.len() % 4 == 0, "input is not 4-byte-aligned");
+    // concatenate the method selector to the input
+    let input: Vec<u8> = method
+      .unwrap_or(METHOD_SELECTOR_DEFAULT)
+      .bytes()
+      .into_iter()
+      .chain(input.into_iter())
+      .collect();
```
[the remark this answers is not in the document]
poszu: Actually, it might be better to use scale encoding than serde to be consistent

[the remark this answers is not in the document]
lrettig: We can encapsulate the serialize/deserialize logic (at least in Rust) in the struct itself, can't we?

core/src/syscall/host.rs
```diff
@@ -90,49 +88,23 @@ impl Syscall for SyscallHostCall {
...
-          .collect(),
-      )
+      // Round up the length to the nearest word boundary
+      let input_words = ctx.slice(arg2, (len + 3) / 4);
+      let input_bytes = input_words
+        .into_iter()
+        .flat_map(|word| word.to_le_bytes())
+        .collect::<Vec<u8>>();
+      // this removes any extra padding from the input
+      Some(input_bytes[..len].to_vec())
```
poszu, line 102: How about
```suggestion
	.take(len)      // this removes any extra padding from the input
        .collect::<Vec<u8>>();
      Some(input_bytes)
```

to avoid creating 2 vectors.

core/src/syscall/host.rs
```diff
@@ -57,11 +57,9 @@ impl Syscall for SyscallHostWrite {
 /// SyscallHostCall performs a host call, calling other programs.
 /// Inputs:
 ///  - a0 (arg1): address to call
-///  - a1 (arg2): pointer to input (bytes) to pass to the called program
+///  - a1 (arg2): pointer to input (bytes) to pass to the called program (first four bytes are method selector)
```
poszu, line 60:
```suggestion
///  - a1 (arg2): pointer to payload containing method selector and input to the called program
```

core/src/runtime/mod.rs
```diff
@@ -715,6 +716,22 @@ impl<'host> Runtime<'host> {
     self.execute()
   }
 
+  /// Execute an exported function using method selector. Does the same work as execute().
+  pub fn execute_selector(
```
poszu, line 720: This sounds weird. It executes the selected method, not the selector :D
```suggestion
  pub fn execute_method(
```

lrettig: I made the names more explicit

interface/Cargo.toml
```diff
@@ -8,6 +8,10 @@ license.workspace = true
 edition.workspace = true
 
 [dependencies]
+bincode = "1.3.3"
 blake3 = "1.5.3"
 bytemuck = "1.16.0"
+hex = "0.4.3"
+parity-scale-codec = { version = "3.6.12", features = ["derive"] }
+serde = { version = "1.0.204", features = ["derive"] }
```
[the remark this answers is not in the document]
lrettig: Serialize/Deserialize are required for `MethodSelector` so that it can be used as a key in the symbol table BTreeMap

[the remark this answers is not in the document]
poszu: Right, but I wonder why `athena_core::runtime::Program` derives `Serialize` and `Deserialize` in the first place. AFAIR, it doesn't need to. What is the use case for serializing a `Program`? IMO, it could be removed.

poszu (approved):

interface/Cargo.toml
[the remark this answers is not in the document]
lrettig: Gotcha. Those have always been there, but it looks like we don't need them. I removed them. Likely due to a feature in SP1 (continuations?) that we don't have or removed.

PR #157, opened 2024-10-16
Fix and test wallet `verify` method

### LINKED ISSUE #129: API: Verify
Given a transaction, go-spacemesh needs to know whether the transaction is valid. More specifically, given context including `principal` Account and `principal_template` Account and a full, raw tx, this method needs to call the `Verify()` method on the principal template, pass in the raw tx (including the signature), and return the bool result to go-spacemesh. The single sig wallet template should simply check that the ed25519 signature on the tx matches the pubkey stored in the principal account.

poszu:
Closes #129 

The example wallet (`examples/wallet/program`) now properly implements `VerifiableTemplate::verify()`. This method is executable just like every other program method via calling `athexp_verify` and providing the required set of inputs:
- wallet state,
- tx bytes as vector of bytes,
- ts signature as 64B array.

I also added tests to check signature verification in:
- example/wallet/script,
- Go bindings (calling Verify from host, over FFI).


vm/sdk/src/lib.rs
```diff
@@ -49,11 +49,20 @@ pub(crate) fn bytes_to_u32_vec<T: AsRef<[u8]>>(bytes: T) -> Vec<u32> {
 }
 
 pub trait VerifiableTemplate {
-  fn verify(&self, tx: &[u8], signature: &[u8; 64]) -> bool;
+  fn verify(&self, tx: Vec<u8>, signature: [u8; 64]) -> bool;
+}
+
+pub fn encode_verify(state: Vec<u8>, tx: &[u8], signature: &[u8; 64]) -> Vec<u8> {
```
lrettig, line 55: I don't think we need an `encode_verify`. The host needs to call verify when it has a raw tx, but there's no such thing as a "verify tx". When calling verify, the host doesn't need to decode any of the tx contents. It can just pass in the entire payload.

lrettig (changes requested):

examples/wallet/program/src/main.rs
```diff
@@ -56,32 +56,20 @@ impl WalletProgram for Wallet {
...
-    let wallet = Wallet::new(owner);
-
-    Wallet::spawn(owner);
-    let send_arguments = SpendArguments {
-      recipient: [0u8; 24],
-      amount: 0,
-    };
-    wallet.spend(send_arguments);
-  }
+#[link_section = ".text.athexp.verify"]
```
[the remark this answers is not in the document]
poszu: Because it requires rethinking and probably many changes in the macro. The proc macro needs to be placed on `impl WalletProgram for Wallet` and it cannot be placed in the same file twice because it would cause all sorts of errors like registering a global allocator twice, etc. I don't want to mix it in this PR - it can be done later.

vm/sdk/src/lib.rs
poszu: It exists to let the host encode all the required information (account state, raw tx and its signature) into a single vector of bytes to be assigned to `msg.input_data` when it calls `vm.execute(.., msg)`. The use case I've been thinking when it is needed:

1. Host receives a signed transaction (`tx []byte, signature [64]byte`)
2. It loads the wallet state
3. It prepares `C.athcon_message` by assigning  `msg.input_data = athcon_encode_verify(state, tx, signature)`. Now `msg.input_data` contains `method_selector("athexp_verify") | state | tx | signature`.
4. It calls `C.athcon_execute(VM, msg)`
5. The VM executes `wallet.verify(state, tx, signature)` (state == &self)

lrettig: As discussed elsewhere, in order to avoid the unnecessary roundtrip described in point 3., especially given that the host _already_ has the encoded state, I suggest that we add another top-level data structure here:

ExtendedPayload := {account_state_blob byte[], ExecutionPayload}
ExecutionPayload := {MethodSelector, args byte[]}

Then the host can generate ExtendedPayload without needing to talk to the VM. ExtendedPayload here is the data structure that gets encoded and passed in as `msg.input_data`.

interface/src/lib.rs
```diff
@@ -877,6 +878,8 @@ mod tests {
 
   #[test]
   fn test_method_selector() {
+    let selector = MethodSelector::from("athexp_verify");
+    dbg!(format!("{selector}"));
```
poszu, line 882:
```suggestion
```

vm/sdk/src/lib.rs
```diff
@@ -50,7 +50,7 @@ pub(crate) fn bytes_to_u32_vec<T: AsRef<[u8]>>(bytes: T) -> Vec<u32> {
 }
 
 pub trait VerifiableTemplate {
-  fn verify(&self, tx: &[u8], signature: &[u8; 64]) -> bool;
+  fn verify(&self, tx: Vec<u8>, signature: [u8; 64]) -> bool;
```
lrettig, line 53: Why change this function signature?

lrettig (approved):

vm/sdk/src/lib.rs
poszu: Because the decoding from io needs an owned type to decode into. It cannot decode into a reference type.

PR #159, opened 2024-10-17
remove example programs' elfs

poszu:
The ELFs included in the repo are binary files and cannot be versioned "as a diff". Git stores a copy of every elf in history and the repository grows rapidly. 

The binaries are needed for some tests, which execute guest programs. In theory they shouldn't change a lot, but in practice, most of them change every PR. In my opinion, it's better to build the binaries automatically on demand (via build.rs).

This PR:
- removes all ELF binaries from the repo,
- adds build.rs files to `core` and `runner`, where tests need the binaries

Little explanation is needed about how the build.rs files work. We don't want to build the binaries every time when a crate is compiled, but only for tests. A `build.rs` is missing an easy way to detect that cargo is building tests. To detect it, I added a feature flag `unittests` which is enabled for tests via a self-pointing dev-dependency:

```toml
[features]
unittest = []

[dev-dependencies]
athena-core = { path = ".", features = ["unittest"] }
```


.github/workflows/pr.yml
```diff
@@ -11,6 +11,40 @@ concurrency:
...
+          curl -L https://install.athenavm.org | bash
+          $ATHENA_DIR/bin/athup --path $GITHUB_WORKSPACE
+          $ATHENA_DIR/bin/cargo-athena athena --version
+          echo "$ATHENA_DIR/bin" >> $GITHUB_PATH
+
+      - name: Install Athena CLI
+        run: |
+          cd cli
+          cargo install --force --locked --path .
+          cargo athena install-toolchain
```
lrettig, line 33: Pretty sure the `athup` command above installs the toolchain. I think you don't need both of these steps.

.github/workflows/pr.yml
```diff
@@ -11,6 +11,40 @@ concurrency:
...
+
+      - name: Build binaries
+        run: |
+          make -C tests
+          make -C examples
+
+      - name: Upload host-test binary
+        uses: actions/upload-artifact@v4
```
lrettig, line 41: Is it worth caching these?

.gitignore
```diff
@@ -19,3 +19,4 @@ target/
 
 # Still not sure why cargo creates this in the root directory
 /elf
+**/elf/*
```
lrettig, line 22: Why not replace the previous line and this line with `elf/` ?

.github/workflows/pr.yml
```diff
@@ -31,6 +65,19 @@ jobs:
...
+          $ATHENA_DIR/bin/cargo-athena athena --version
+          echo "$ATHENA_DIR/bin" >> $GITHUB_PATH
+
+      - name: Install Athena CLI
+        run: |
+          cd cli
+          cargo install --force --locked --path .
+          cargo athena install-toolchain
+
       - name: Run cargo check
```
lrettig, line 81: Running `cargo check` here actually rebuilds the binaries: https://github.com/athenavm/athena/actions/runs/11391768813/job/31696289475?pr=159

Maybe the `build-binaries` job isn't necessary?

lrettig (changes requested):

.github/workflows/pr.yml
poszu: I think it doesnt. cargo install is required for cargo athena to be available.

.github/workflows/pr.yml
poszu: I added build binaries job to avoid installing athena in every job, where it is normally not needed otherwise. I takes quite a time to install athena and build all binaries so it also speeds up CI.

.github/workflows/pr.yml
poszu: Yes, I think so. Installation of the toolchain and building all tests and examples takes significant time. I've also seen toolchain installation fail because of exhausted HTTP query limits in the past. Installing it fewer times should help avoid this.

poszu: Btw, it still happens unfortunately: https://github.com/athenavm/athena/actions/runs/11440975992/job/31829075665?pr=159

.github/workflows/pr.yml
lrettig: don't you love that friendly, clear, helpful error message? :)

.github/workflows/pr.yml
lrettig: don't you love that friendly, clear, helpful error message? :)

.github/workflows/pr.yml
lrettig: If you just run `athup` in default mode, it both sets up `~/.cargo/bin/cargo-athena` _and_ sets up the toolchain. If you run it with `--path` as we do here, it does the former (with a link to the local build) but not the latter. (Do we want to change this?) So, you need to run `cargo athena install-toolchain` but you do _not_ need to run `cargo install --force --locked --path cli` -- I just double checked this. If `cargo athena` isn't working after the athup command, it must be a PATH issue.

.github/workflows/pr.yml
lrettig: I'm confused. This job also includes "Install Athena toolchain" and "Install Athena CLI". How is this saving time, the way you have this set up? Installing the toolchain is relatively slow since it's a big file, but, to reiterate, running `cargo check` here is going to rebuild the binaries _anyway._

poszu: Yes, `cargo check` will rebuild binaries in this job, but this job specifically does not use the artifacts from `build-binaries`. Its artifacts are used in `test-go` and `lint` jobs.

.github/workflows/pr.yml
```diff
@@ -65,30 +114,26 @@ jobs:
...
-      - name: Setup CI
-        uses: ./.github/actions/setup
+      - name: Setup Rust toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1
 
       - name: Run cargo fmt
-        uses: actions-rs/cargo@v1
+        run: cargo fmt --all -- --check
+
+      - name: Download binaries
```
lrettig, line 130:
```suggestion
      - name: Download test binaries
```

.github/workflows/pr.yml
```diff
@@ -19,42 +53,57 @@ jobs:
...
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+
+      - name: Install Athena toolchain
+        run: |
+          curl -L https://install.athenavm.org | bash
+          $ATHENA_DIR/bin/athup --path $GITHUB_WORKSPACE
+          $ATHENA_DIR/bin/cargo-athena athena --version
+          echo "$ATHENA_DIR/bin" >> $GITHUB_PATH
+
+      - name: Install Athena CLI
```
lrettig, line 75: As mentioned above: you don't need to _both_ install and run athup, and also run `cargo install cli`. Remove one or the other. If the goal here is to run the _local_ version of the CLI, let's drop "Install Athena toolchain" in this job entirely.

.gitignore
lrettig: this doesn't appear to have changed, so, same question

lrettig (changes requested):

.gitignore
poszu: Changed to `elf/`

.github/workflows/pr.yml
poszu: Fixed by removing the `athup` step. Now it is installing `cli` to have `cargo athena` available (needed for `make -C tests/examples`) and the toolchain (to actually build the binaries).

.github/workflows/pr.yml
poszu: fixed

lrettig (approved):

Issue #92, opened 2024-08-15
Add logs


lrettig:
We need EVM-style logs, i.e., a running program needs a way to log the fact that a particular event occurred (coins sent, program spawned, etc.).

poszu:
@lrettig do you mean regular logs that are printed to stdout or something more sophisticated? The programs can already write to stdout by using the WRITE syscall with the stdout file descriptor. E.g, `println!()` works like this and I assume that `log::info!()` and others with a subscriber configured to print on stdout would work as well.

lrettig:
No, something more sophisticated. I know the name "log" is misleading, but I'm envisioning something more like how logs/events work in EVM: https://medium.com/mycrypto/understanding-event-logs-on-the-ethereum-blockchain-f4ae7ba50378

EVM has opcodes for emitting logs, and they're actual data committed to the blockchain and that all nodes achieve consensus on. They're stored in a log trie that's part of the account state trie.

We don't have to do the exact same thing, but at the very least we definitely want to support some kind of events that both user-facing applications, as well as backend infrastructure (e.g., explorers, data indexers) can subscribe to.

lrettig:
yes, they're specific to a program, but we also want standards (like ERC-20) so, e.g., all token programs use the same log messages to make indexing easy. The more interesting question here isn't the opcode we use but how these data are surfaced to the application layer (which we probably need to handle on the go-sm side, e.g., via API changes)

Issue #36, opened 2024-07-02
Call return values


lrettig:
To keep things simple, the `call` syscall was implemented without support for a return value (#5). We probably want to add additional syscalls with signatures like:

```rust
pub fn set_return_value(buf: *const u8, length: u64);
pub fn get_return_value(buf: *mut u8, length: u64);
```

EVM achieves this using the `RETURN` and `REVERT` opcodes. Solana has syscalls that are very similar to these. Here are some helpful resources:

- https://docs.solanalabs.com/proposals/return-data
- https://book.anchor-lang.com/anchor_in_depth/CPIs.html#returning-values-from-handler-functions
- https://docs.soliditylang.org/en/v0.8.26/units-and-global-variables.html#members-of-address-types

PR #169, opened 2024-10-22
remove unused ExecutionResult::create_address

poszu:
The address of the spawned program is returned the same way as the results of all contract methods via the `ExecutionResult::output`.


lrettig:
A few questions:

1. What happens when _multiple_ spawns occur in a single call, or a single program execution? E.g., imagine a program serially spawning multiple other programs.
2. What happens when a spawn occurs lower down in the call stack? E.g., imagine a program makes an external call that results in a spawn. How does it get the result of that spawn?
3. What happens if both of these occur, i.e., a program makes an external call that results in multiple spawns?
4. What about the case described in this comment?:

https://github.com/athenavm/athena/blob/46a70843893faa481f1af380a8ea9f279656a9a0/ffi/athcon/include/athcon/athcon.h#L394-L403

This is a nuance involving gas pricing and the cost of state access. In a single transaction, we need to consider the fact that the first time an account is touched (or created), the cost is higher than subsequent accesses (this is what is meant by "warm"). We *probably* don't need to worry about this for now, but I just want to make sure that we're aware of it.

poszu:
Let me give it more thought. I think we could return the spawned addresses in every case "normally" by bubbling it up the call stack, if the `CALL` syscall supported returning results (#36). For example,

>1. What happens when multiple spawns occur in a single call, or a single program execution? E.g., imagine a program serially spawning multiple other programs.

This program could return `Vec<Address>`, which would be available to the caller in _results_.

>2. What happens when a spawn occurs lower down in the call stack? E.g., imagine a program makes an external call that results in a spawn. How does it get the result of that spawn?
>3. What happens if both of these occur, i.e., a program makes an external call that results in multiple spawns?

The address(es) could be simply returned from the `CALL` syscall supported returning values (#36)

>4. What about the case described in this comment?:

I don't understand the use case described in the comment. What address would it return if spawning failed? Anyhow, probably the same method could be used (return the address in execution results).

lrettig:
> > 1. What happens when multiple spawns occur in a single call, or a single program execution? E.g., imagine a program serially spawning multiple other programs.
> 
> This program could return `Vec<Address>`, which would be available to the caller in _results_.

Only one thing to add: the result of each `spawn` can already be available to the program performing the spawn, as each spawn host call returns the newly-created contract address. The question of how to surface this to the user is a bit more complex. The other obvious way to do this, aside from a program return value, is logs/events (#92).

lrettig (approved):

PR #170, opened 2024-10-22
cleanup release workflow

### LINKED ISSUE #184: bug: release workflow failed
The release workflow has failed. Some or all binaries might have not been published correctly.

Check the [release workflow page](https://github.com/athenavm/athena/actions/runs/11555852321
) for details.

This issue was raised by the workflow at `.github/workflows/release.yml`.

poszu:
Fixes #184 
- removed "setup ci" action, which contained many steps that weren't needed in most jobs (i.e. installing Go for rust tests and vice versa)
- removed unsupported macos-13
- cleaned up release workflow a bit

An example of passed release workflow: https://github.com/athenavm/athena/actions/runs/11460107106 which resulted in a nightly release: https://github.com/athenavm/athena/releases/tag/nightly-6010924ab2fd9362060fd26ae6748663914f9f15


.github/actions/setup/action.yml
```diff
@@ -1,30 +0,0 @@
-name: Test setup
-inputs:
-  pull_token:
-    description: "Token to use for private repo access"
-    required: true
-runs:
-  using: "composite"
-  steps:
-    - name: Set up git private repo access
```
lrettig, line 9: Are we sure we want to remove this? I thought the idea here was to avoid some of the github rate limiting issues we run into when installing the athena toolchain

poszu: This is not used. The PULL_TOKEN is not even configured in the repo settings and if it was, it would only be passed in the release workflow (not in pr). I suspect it was needed in the repo it was copied from for pulling from private repos (and ours are public).

lrettig (approved):

lrettig:
It would be nice to be able to switch on "static mode" for the VM, which basically just ensures that no state writes occur/that there are no side effects from a program execution. For now I think it's sufficient to make sure that in static mode we don't use certain host calls, such as write_storage and spawn. Ideally this could be switched on for a single execute() call, and it should be persistent: when switched on, _all_ subsequent stack frames should be in static mode until we return to the place where it was switched on.

One place this would be useful is when the host calls certain methods, including Verify and MaxSpend, which it currently assumes are static--it would be good to have additional assurances of this from the VM.

It's also likely we'll want to introduce something like EVM's [`STATICCALL`](https://www.rareskills.io/post/solidity-staticcall) at some point.

PR #173, opened 2024-10-22
Improve error handling in Go bindings

lrettig:
Bring in the rest of the ATHCON errors. Allow athena errors in Go to wrap underlying Go errors.

These are quality of life improvements that make the go-sm integration work easier.


ffi/athcon/bindings/go/athcon.go
```diff
@@ -192,7 +222,8 @@ func (vm *VM) Execute(
 	code []byte,
 ) (res Result, err error) {
 	if len(code) == 0 {
-		return res, fmt.Errorf("code is empty")
+		fmt.Fprintf(os.Stderr, "athcon: no input code")
+		return res, Failure
```
poszu, line 226: I'd prefer it not to print anything and leave it to the lib user. How about adding the context in the returned error? Like: 
```suggestion
		return res, fmt.Errorf("athcon: %w: no input code", Failure)
```

ffi/athcon/bindings/go/athcon.go
```diff
@@ -31,19 +32,48 @@ const (
...
+	OutOfGas                  = Error{Code: C.ATHCON_OUT_OF_GAS}
+	InvalidInstruction        = Error{Code: C.ATHCON_INVALID_INSTRUCTION}
+	UndefinedInstruction      = Error{Code: C.ATHCON_UNDEFINED_INSTRUCTION}
+	StackOverflow             = Error{Code: C.ATHCON_STACK_OVERFLOW}
+	StackUnderflow            = Error{Code: C.ATHCON_STACK_UNDERFLOW}
+	BadJumpDestination        = Error{Code: C.ATHCON_BAD_JUMP_DESTINATION}
+	InvalidMemoryAccess       = Error{Code: C.ATHCON_INVALID_MEMORY_ACCESS}
+	CallDepthExceeded         = Error{Code: C.ATHCON_CALL_DEPTH_EXCEEDED}
+	StaticModeViolation       = Error{Code: C.ATHCON_STATIC_MODE_VIOLATION}
+	PrecompileFailure         = Error{Code: C.ATHCON_PRECOMPILE_FAILURE}
```
poszu, line 66: Does it make sense to expose all possible errors to the users of these bindings? I suspect some of them will never be used. For example, the host side would never return `UndefinedInstruction`.

poszu (changes requested):

ffi/athcon/bindings/go/athcon.go
lrettig: I don't see any harm in wrapping them all. In my mind that's sort of the purpose of the bindings, and it should be pretty unopinionated about what the client does with the exposed data types.

poszu: The harm is that it leaks the internal details of the VM to the users. These errors are exported, and any change in them is technically a breaking change. IMO, the rule of thumb is to expose as little as possible.

ffi/athcon/bindings/go/athcon.go
```diff
@@ -192,7 +221,10 @@ func (vm *VM) Execute(
 	code []byte,
 ) (res Result, err error) {
 	if len(code) == 0 {
-		return res, fmt.Errorf("code is empty")
+		return res, Error{
+			Code: C.ATHCON_FAILURE,
+			Err:  fmt.Errorf("athcon execute: no input code"),
```
poszu, line 226:
```suggestion
			Err:  errors.New("athcon execute: no input code"),
```

poszu (approved):

poszu (approved):

PR #175, opened 2024-10-23
CALL syscall returns data

### LINKED ISSUE #36: Call return values

poszu:
Add functionality of returning data from the program executed via a `CALL` syscall. The syscall helper supports returning up to 1024B for now. It allocates a buffer for the (possible) returned data on the caller's heap, which is populated (if needed) by the VM syscall handler (if the `call` returns any data).

Closes #36 

Other changes:
- updated the recursive-call test to use the data returned from the `call()` instead of the host storage.
- refactored the `recursive-call-test` to be included in the main workspace (similarly as the `host-test` before). It gets linting, formatting, go-to definition and so on. Its tests are automatically executed with `cargo t`.


tests/host/src/main.rs
lrettig, line 1: Line 47: `// Note: there is no way to check the result of a call`

I think this comment needs to be updated

tests/recursive_call/src/main.rs
```diff
@@ -1,56 +1,36 @@
...
 use athena_vm_sdk::call;
 
-// Note: the test harness installs this contract code at ADDRESS_ALICE
-
-// emulate a return value by writing the return value to storage.
-// Athena doesn't support return values yet.
 fn return_value(value: u32) {
-  let key = bytes32_to_32bit_words(STORAGE_KEY);
-  let val: [u32; 8] = [value, 0, 0, 0, 0, 0, 0, 0];
-  let status = write_storage(&key, &val);
```
lrettig, line 18: It's nice that this test is now much cleaner, but can we make sure we have another test that tests writing to and reading from storage instead?

tests/recursive_call/tests/test.rs
```diff
@@ -0,0 +1,27 @@
...
+fn test() {
+  tracing_subscriber::fmt::init();
+
+  let elf = include_bytes!("../elf/recursive-call-test");
+  let mut stdin = AthenaStdin::new();
+  let vm = AthenaVm::new();
+  let mut host = MockHost::new_with_vm(&vm);
+  let template_address = [0x77; 24];
+  stdin.write(&(template_address, 6u32));
+  //   stdin.write(&);
```
lrettig, line 15: remove this

vm/entrypoint/src/syscalls/host.rs
```diff
@@ -13,17 +13,29 @@ use athena_interface::StorageStatus;
...
         in("a1") input_ptr,
         in("a2") input_len,
-        in("a3") amount,
-    )
+        in("a3") output_ptr,
+        in("a4") output_size,
+        in("a5") amount,
+        lateout("t0") size,
+    );
+    return size;
```
lrettig, line 38:
```suggestion
    size
```

vm/sdk/src/call.rs
```diff
@@ -12,5 +20,25 @@ pub fn call(address: Address, input: Option<Vec<u8>>, method: Option<&str>, amou
...
+    input.as_ptr(),
+    input_len,
+    output_buf.as_mut_ptr() as *mut u32,
+    1024,
+    amount.as_ptr(),
+  );
+
+  // Convert the Vec<MaybeUninit<u32>> into a Vec<u8> without freeing buffer
+  let raw_ptr = output_buf.as_mut_ptr() as *mut u8;
+  let capacity = output_buf.capacity();
```
lrettig, line 38: I think this is wrong, `output_buf.capacity()` here is 256, but this should be in bytes not words

lrettig:
```suggestion
  let capacity = output_buf.capacity() * std::mem::size_of::<u32>();
```

vm/sdk/src/call.rs
```diff
@@ -12,5 +20,25 @@ pub fn call(address: Address, input: Option<Vec<u8>>, method: Option<&str>, amou
...
+
+  // Allocate 1024 bytes on the heap, using u32 for proper alignment
+  let mut output_buf = Vec::<std::mem::MaybeUninit<u32>>::with_capacity(1024 / 4);
+
+  let output_size = athena_vm::syscalls::call(
+    address.as_ptr(),
+    input.as_ptr(),
+    input_len,
+    output_buf.as_mut_ptr() as *mut u32,
+    1024,
```
lrettig, line 32:
```suggestion
    1024, // TODO: may be changed, see https://github.com/athenavm/athena/issues/177
```

lrettig: Considering this value is used in multiple places, can we make this a const and define it only once? E.g., it's also used on line 25 above

core/src/syscall/host.rs
```diff
@@ -123,6 +127,30 @@ impl Syscall for SyscallHostCall {
     let host = ctx.rt.host.as_deref_mut().expect("Missing host interface");
     let res = host.call(msg);
 
+    let output_size = if let Some(output) = res.output {
+      let mut output_ptr = ctx.rt.register(Register::X13);
+      let output_size = ctx.rt.register(Register::X14);
+      let output_size = min(output.len() as u32, output_size);
```
lrettig, line 133: Add a warning if `output.len() > output_size` so it's clear that output is being truncated

vm/sdk/src/call.rs
```diff
@@ -12,5 +20,25 @@ pub fn call(address: Address, input: Option<Vec<u8>>, method: Option<&str>, amou
...
+    amount.as_ptr(),
+  );
+
+  // Convert the Vec<MaybeUninit<u32>> into a Vec<u8> without freeing buffer
+  let raw_ptr = output_buf.as_mut_ptr() as *mut u8;
+  let capacity = output_buf.capacity();
+  // Prevent `buffer` from being dropped, as we're transferring ownership
+  std::mem::forget(output_buf);
```
lrettig, line 40: consider this instead, since the intent is a bit clearer:

```
  // Prevent `buffer` from being dropped, as we're transferring ownership
  let mut output_buf = std::mem::ManuallyDrop::new(output_buf);
```

You could remove this line entirely, and add the new line above, before using `output_buf` after the sys call.

vm/sdk/src/call.rs
```diff
@@ -12,5 +20,25 @@ pub fn call(address: Address, input: Option<Vec<u8>>, method: Option<&str>, amou
...
+
+  // Convert the buffer into Vec<u8>, accounting for the number of bytes written
+  unsafe { Vec::from_raw_parts(raw_ptr, output_size, capacity) }
```
lrettig, line 43: It couldn't hurt to add an assert here that `output_size <= capacity`. Your host call implementation doesn't allow this, but it's nice to be explicit when using unsafe code to prevent confusing memory errors down the line if anything changes

lrettig (changes requested):

vm/entrypoint/src/syscalls/host.rs
poszu: It must return here to not execute `unreachable!`.

vm/sdk/src/call.rs
poszu: Done

tests/recursive_call/src/main.rs
poszu: The tests in https://github.com/athenavm/athena/blob/main/tests/host/src/main.rs cover it already :)

vm/sdk/src/call.rs
```diff
@@ -1,16 +1,41 @@
...
     selector: method.map(MethodSelector::from),
     input: input.unwrap_or_default(),
   };
   let input_payload = payload.encode();
-  let input_len = input_payload.len();
-  let input = crate::bytes_to_u32_vec(input_payload);
-  athena_vm::syscalls::call(address.as_ptr(), input.as_ptr(), input_len, amount.as_ptr());
+
+  // Allocate output buffer on the heap, aligned to 4B.
+  // It's capacity might be configureable and changed in the future.
```
lrettig, line 21:
```suggestion
  // Its capacity might be configurable and changed in the future.
```

lrettig (approved):

PR #180, opened 2024-10-24
Add another verify test

lrettig:
This more closely mimics the behavior of receiving a verify tx over FFI. Also adds some more debug output.

In particular, this handles serialization and deserialization a bit differently. The existing test serializes and adds to the stdin stream each input object separately. This first encodes the payload, then serializes it and adds it to the stdin stream as a single blob.


vm/sdk/src/wallet.rs
```diff
@@ -11,6 +11,17 @@ pub struct SpendArguments {
...
+#[derive(Clone, Debug, Encode, Decode, PartialEq, Eq)]
+pub struct VerifyArguments {
+  pub raw: Vec<u8>,
+  pub sig: [u8; 64],
+}
+
+pub fn encode_verify_inner(raw: Vec<u8>, sig: [u8; 64]) -> Vec<u8> {
+  let args = VerifyArguments { raw, sig };
+  args.encode()
+}
```
poszu, line 23: Two comments. 

From the perspective of SCALE encoding, the `VerifyArguments` structure is equivalent to a tuple `(Vec<u8>, [u8; 64])` so a custom type to be used only in this one method isn't needed. This is equivalent: 
```suggestion
pub fn encode_verify_inner(raw: Vec<u8>, sig: [u8; 64]) -> Vec<u8> {
  (raw, sig).encode()
}
```

Another thing is that it is weird to have a public `*_inner()` function.

PR #182, opened 2024-10-25
Precompile syscall for ed25519 signature verification

### LINKED ISSUE #179: Add ed25519 precompile
A single verify() call consumes > 10M cycles and takes ~20 seconds on my PC. It looks like we may need to start adding precompiles sooner rather than later, or the testnet is going to be very slow.

We can probably take this directly from sp1: https://github.com/succinctlabs/sp1/tree/dev/crates/core/executor/src/syscalls/precompiles/edwards

poszu:
Closes #179 

Added a new syscall for verifying ED25519 signatures. The new syscall has a dummy cost of 100 cycles (left a todo to update it in the future). The test which verifies TX signatures now runs with much less gas and faster.


core/src/runtime/syscall.rs
```diff
@@ -47,6 +51,7 @@ impl SyscallCode {
     match value {
       0x00_00_00_00 => SyscallCode::HALT,
       0x00_00_00_02 => SyscallCode::WRITE,
+      0x00_64_00_20 => SyscallCode::PRECOMPILE_ED25519_VERIFY,
```
lrettig, line 54: let's keep these in order... there will be more precompiles, let's group them together below the existing sys calls

core/src/syscall/precompiles/ed25519.rs
```diff
@@ -0,0 +1,37 @@
...
+    if pubkey.verify(&msg, &signature).is_ok() {
+      tracing::debug!("signature is valid");
+      return Ok(Outcome::Result(Some(1)));
+    }
+    tracing::debug!("signature is invalid");
+    Ok(Outcome::Result(Some(0)))
+  }
+
+  fn num_extra_cycles(&self) -> u32 {
+    // TODO: decide the cost
```
lrettig, line 34:
```suggestion
    // TODO: decide the cost, see https://github.com/athenavm/athena/issues/176
```

examples/wallet/script/src/bin/execute.rs
```diff
@@ -199,7 +199,7 @@ mod tests {
         &MethodSelector::from("athexp_verify"),
         stdin.clone(),
         Some(&mut host),
-        Some(25000000),
+        Some(20000),
```
lrettig, line 202: 🫠

lrettig (changes requested):
love how short this is

core/src/runtime/syscall.rs
poszu: What do you mean? They are ordered by syscall ID. Btw, I realized that each byte of the syscall code has a special meaning (hence 0x64) for 100 gas. I'm not sure if we want to keep it. It's weird to have gas cost encoded in the syscall code.

lrettig: I see. Yeah, afaict this is NOT dictated by the RISC-V spec itself, it seems to be something we carried over from SP1. And it's not enforced, other than with a test. I think we can drop it, but let's do so under a separate PR.

I can't see any benefit to encoding this in the syscall ID, can you?

lrettig (approved):

PR #207, opened 2024-10-31
newtype for athena Address

poszu:
Change `athena_interface::Address` to be a newtype wrapping a `[u8;24]` instead of a type alias. 

Benefits:
- stronger type safety,
- allows for implementing foreign traits (like `std::fmt::Display`, `From`)
- removed some unnecessary wrapper code (`AddressWrapper`)

### DIFF
--- a/ffi/athcon/bindings/rust/athcon-client/tests/integration_tests.rs
+++ b/ffi/athcon/bindings/rust/athcon-client/tests/integration_tests.rs
@@ -8,7 +8,7 @@ use athcon_client::{
 };
 use athcon_vm::{MessageKind, Revision, StatusCode, StorageStatus};
 use athena_interface::payload::ExecutionPayload;
-use athena_interface::ADDRESS_ALICE;
+const ADDRESS_ALICE: Address = [0xAAu8; 24];
 
 const CONTRACT_CODE: &[u8] =
   include_bytes!("../../../../../../tests/recursive_call/elf/recursive-call-test");
--- a/interface/src/lib.rs
+++ b/interface/src/lib.rs
@@ -54,7 +54,31 @@ impl std::fmt::Display for MethodSelector {
   Deserialize,
   Serialize,
 )]
-pub struct Address(pub [u8; ADDRESS_LENGTH]);
+pub struct Address([u8; ADDRESS_LENGTH]);
+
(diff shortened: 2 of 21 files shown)


interface/src/lib.rs
```diff
@@ -337,11 +331,11 @@ pub fn calculate_address(
 ) -> Address {
   // calculate address by hashing the template, blob, principal, and nonce
   let mut hasher = Hasher::new();
-  hasher.update(template);
+  hasher.update(&template.0);
```
lrettig, line 334: wouldn't an Into method for a byte slice make this more readable?

ffi/athcon/bindings/rust/athcon-client/tests/integration_tests.rs
```diff
@@ -8,7 +8,7 @@ use athcon_client::{
 };
 use athcon_vm::{MessageKind, Revision, StatusCode, StorageStatus};
 use athena_interface::payload::ExecutionPayload;
-use athena_interface::ADDRESS_ALICE;
+const ADDRESS_ALICE: Address = [0xAAu8; 24];
```
lrettig, line 11: Why redefine this here?

vm/entrypoint/src/helpers/mod.rs
```diff
@@ -1,20 +1,8 @@
-use athena_interface::{Address, Balance, Bytes32, Bytes32Wrapper};
+use athena_interface::Bytes32;
 use bytemuck::cast;
 
 // Simple helper functions for smart contracts
 
-pub fn address_to_bytes32(address: Address) -> Bytes32 {
```
lrettig, line 6: I see, this is nice

jellonek (approved):

ffi/athcon/bindings/rust/athcon-client/tests/integration_tests.rs
poszu: I'm slowly removing all uses of the dummy addresses defined in the `athena-interface` with the end goal of removing them entirely. Every test should define the addresses it wants to work with itself.

interface/src/lib.rs
poszu: `Into` consumes the value.  I will add implementations of `AsRef` instead.

Issue #215, opened 2024-11-05
symbol not found in flat namespace '__end'


lrettig:
#161 is back again, this time on the execution side (and only on mac). From https://github.com/spacemeshos/go-spacemesh/actions/runs/11674170614/job/32506374469?pr=6380:

```
=== FAIL: node TestSpacemeshApp_TransactionService (0.49s)
panic: loading Athena VM: loading library: dlopen(/Users/m1/actions-runner/_work/go-spacemesh/go-spacemesh/build/libathenavmwrapper.dylib, 0x000A): symbol not found in flat namespace '__end' [recovered]
	panic: loading Athena VM: loading library: dlopen(/Users/m1/actions-runner/_work/go-spacemesh/go-spacemesh/build/libathenavmwrapper.dylib, 0x000A): symbol not found in flat namespace '__end'

goroutine 1817 [running]:
testing.tRunner.func1.2({0x106fe5ce0, 0xc0010a4280})
	/Users/m1/actions-runner/_work/_tool/go/1.23.2/arm64/src/testing/testing.go:1632 +0x2c4
testing.tRunner.func1()
	/Users/m1/actions-runner/_work/_tool/go/1.23.2/arm64/src/testing/testing.go:1635 +0x47c
panic({0x106fe5ce0?, 0xc0010a4280?})
	/Users/m1/actions-runner/_work/_tool/go/1.23.2/arm64/src/runtime/panic.go:785 +0x124
github.com/spacemeshos/go-spacemesh/vm/sdk/wallet.Spawn({0xc00059ecc0, 0x40, 0x40}, 0x0, {0xc0006cfe70, 0x1, 0x0?})
	/Users/m1/actions-runner/_work/go-spacemesh/go-spacemesh/vm/sdk/wallet/tx.go:45 +0x5d0
github.com/spacemeshos/go-spacemesh/node.TestSpacemeshApp_TransactionService(0xc00029b[380](https://github.com/spacemeshos/go-spacemesh/actions/runs/11674170614/job/32506374469?pr=6380#step:14:381))
	/Users/m1/actions-runner/_work/go-spacemesh/go-spacemesh/node/node_test.go:554 +0xc98
testing.tRunner(0xc00029b380, 0x1072b89f8)
	/Users/m1/actions-runner/_work/_tool/go/1.23.2/arm64/src/testing/testing.go:1690 +0x188
created by testing.(*T).Run in goroutine 1
[...]

PR #217, opened 2024-11-05
Platform-independent heap position

lrettig:
Addresses the issue in #215. It compiles on both Linux and macOS and all tests pass.

Still, I want to investigate the issue a bit further. I don't love hardcoding a platform-independent value here.


poszu:
The code where `_end` is used is a custom memory allocator used only in the guest programs. It's never used and should probably not even be compiled for the VM library. I'd go a slightly different route and:
1. compile it only when building guest programs
2. use a global static byte array for the heap instead of hacking with pointers

Wdyt?

PR #233, opened 2024-11-19
remove nonce and balance from the Wallet example state

poszu:
Following the discussion on Slack https://spacemesh.slack.com/archives/C060PGXV27M/p1731994041096179?thread_ts=1731681768.842079&cid=C060PGXV27M


lrettig (approved):

lrettig:
For context: these were added by mistake (carried over from the existing spacemesh template code) and should never have been there in the first place. For now, we maintain a strict separation of concerns where the host (go-spacemesh) is responsible for managing balance and nonce. A running program must access these using host methods.

PR #234, opened 2024-11-20
release the lib on windows

poszu:
Enable testing the Go bindings on all supported targets and release the library for windows. 

A successful pre-release with windows artifacts: https://github.com/athenavm/athena/releases/tag/nightly-70faee4e6d3c2dbf3b6325732ae767b2c9e744e3

The library on windows exhibited linker issues similar to #215. I found that this happens because the code, which is meant for running in the VM (the contents of `athena-vm-sdk` and `athena-vm` crates, specifically the `athena-vm::heap` package which contains the heap allocator for the guest programs) is a dependency of the `athena-vmlib` crate. This dependency was added to add export the encoding functions for the wallet. 

I temporarily (?) resolved this problem by making the `athena-vm` an optional dependency of the `athena-vm-sdk` (enabled by default) to allow the `athena-vmlib` opting out from importing the VM-side code.

Long term, I think we should rethink whether the `athena-vm-sdk` is the right place for the wallet example structures definitions (like `SpendArguments` etc.). I believe this is not the right place and it should be defined in the wallet crate but I didn't want to refactor it in this PR.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -79,7 +79,14 @@ jobs:
 
   test-go:
     name: Test (Go)
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        runner:
+          - ubuntu-latest
+          - [runs-on, runner=8cpu-linux-arm64, "run-id=${{ github.run_id }}"]
+          - macos-latest
+          - windows-latest
+    runs-on: ${{ matrix.runner }}
     needs: build-binaries
     steps:
       - name: Checkout sources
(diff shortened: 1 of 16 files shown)


jellonek (approved):
For me it looks like this change, especially in [vm/sdk/src/lib.rs](https://github.com/athenavm/athena/pull/234/files#diff-1a9c17ef3b6a70254124f54ee4a8f2ceb17a3a7f8ed9570473a03e3246860ba4) is covering also #235. Right?

poszu:
> For me it looks like this change, especially in [vm/sdk/src/lib.rs](https://github.com/athenavm/athena/pull/234/files#diff-1a9c17ef3b6a70254124f54ee4a8f2ceb17a3a7f8ed9570473a03e3246860ba4) is covering also #235. Right?

Yes, it _fixes_ this problem, but I don't think it's the final solution.

PR #304, opened 2025-01-07
limit the number of VM regs to 16 (RV32e)

### LINKED ISSUE #18: Limit VM registers to 16 for RV32E

poszu:
Part of #18.

Removed registers X16+ from the VM and changed how they are represented in the VM from memory-mapped at "addresses" 0-15 to a separate structure that is also responsible for keeping invariants like the X0 always zero.

Note that, in many places in the VM, a raw `u32` taken from `Instruction` is converted to a `Register`, which may panic if the value is > 15. In the following PR, I will change disassembling logic to convert raw integers to `Register` enum variants at disassembly time to prevent panics at the guest program execution time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 # Athena
 
-Athena is a prototype deterministic smart contract engine that serves as the [Spacemesh network VM][2], and Athena is being designed and built by the  [Spacemesh][8] team. However, Athena is designed to be modular and largely protocol-agnostic so [it will run on other chains][9]. Contributions and integrations are welcome.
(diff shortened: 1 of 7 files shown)


jellonek (approved):

jellonek:
Note: that is partially invalidating the info which is still in README.md about:
> Athena includes a virtual machine (VM) based on the RISC-V ISA, including support for [RV32IM](https://five-embeddev.com/riscv-user-isa-manual/Priv-v1.12/rv32.html#rv32) and [RV32EM](https://five-embeddev.com/riscv-user-isa-manual/Priv-v1.12/rv32e.html),

PR #305, opened 2025-01-08
Refactor Instruction type, catch more errors during transpilation

poszu:
Following up after #304. 

Refactored `Instruction` type to an `enum` with each instruction variant having exactly the operands it needs. Invalid instructions (wrong operand types, registers out of range etc.) are now caught during transpilation instead of runtime and generally, more invariants are guaranteed to be upheld.


core/src/runtime/state.rs
```diff
@@ -7,12 +7,12 @@ use super::Registers;
 /// Holds data describing the current state of a program's execution.
 #[derive(Debug, Clone)]
 pub struct ExecutionState {
-  /// The global clock keeps track of how many instrutions have been executed.
+  /// The global clock keeps track of how many instructions have been executed.
   pub global_clk: u64,
 
-  /// The clock increments by 4 (possibly more in syscalls) for each instruction that has been
+  /// The gas counter increments by 4 (possibly more in syscalls) for each instruction that has been
```
jellonek, line 13: In case of clock I would understand that as instructions can take multiple clock ticks, but why we are keeping the same behavior for gas counter?

jellonek (approved):

core/src/runtime/state.rs
poszu: It's the opposite in our situation. The clock (`global_clk`) ticks once per executed instruction. The gas counter, on the other hand, measures how much "work" has been done already and counts 4 per instruction + an extra cost of some syscalls (these are not "instructions").

jellonek: But why the cost of each instruction is 4 gas points? What is the reason to keep `gas cost == ((number of executed instructions) >> 2) + sum(cost of called syscalls)`?
Why it is 4 instead of (what seems to be more intuitive) 1, or instead of any randomly chosen other number?
That leads to the situation where syscall can cost less than any other instruction.

lrettig: It's an arbitrary number. The exact number doesn't matter; what matters is the cost of each instruction/syscall relative to each of the others. The value 4 seemed reasonable to me; it allows a little bit of flexibility if a "cheaper" instruction type were ever introduced in the future.

PR #320, opened 2025-01-23
simplistic token mint and wallet examples

poszu:
Example contracts implementing a custom token. It features 2 contracts:

### Mint
It allows exchanging SMH for the token, which is then transferred to the chosen wallet address.
                                         
Every mint account instance represents a unique token.

### Token wallet
It keeps and allows spending various tokens created by different mints. 
Every mint instance creates a unique token, which is identified by the address of the mint.
 
---

The contracts are simplistic and could be further extended and improved. For example:
- the mint has a constant exchange ratio (it's part of the immutable state)
- it's not possible to trade tokens back to SMH:w

---

Additionally, this PR extends the `AthenaContext` to keep:
- the caller template, which is needed for the token wallet to identify the caller (it must be either mint or wallet contract)
- the SMH transferred with the call, which is needed for the mint `buy()` to know how much SMH it received

### DIFF
--- a/examples/token/mint/src/main.rs
+++ b/examples/token/mint/src/main.rs
@@ -1,0 +1,78 @@
+#![no_main]
+
+use athena_interface::{Address, Encode, MethodSelector};
+use athena_vm::entrypoint;
+use athena_vm_declare::{callable, template};
+use athena_vm_sdk::{wallet::SpendArguments, Pubkey};
+use parity_scale_codec::Decode;
+
+/// 1000 Smidge == 1 tokens
+const EXCHANGE_RATE: u64 = 1000;
+const SUPPLY_KEY: [u32; 8] = [0; 8];
+
+#[derive(Debug, Decode, Encode)]
+struct Mint {
+  owner: Pubkey,
+  max_supply: u64,
+}
+
+athena_vm::entrypoint!();
+
+#[template]
+impl Mint {
+  #[callable]
+  fn spawn(mint: Mint) -> Address {
+    let mut max_supply = [0u32; 8];
+    max_supply[0] = mint.max_supply as u32;
+    max_supply[1] = (mint.max_supply >> 32) as u32;
+    athena_vm::syscalls::write_storage(&SUPPLY_KEY, &max_supply);
+    athena_vm_sdk::spawn(mint.encode())
+  }
+
+  /// Buy exchanges SMH transfered along with the call into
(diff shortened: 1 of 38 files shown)


examples/token/mint/src/main.rs
```diff
@@ -0,0 +1,103 @@
...
+  #[callable]
+  fn spawn(mint: Mint) -> Address {
+    let address = athena_vm_sdk::spawn(&mint.encode());
+
+    let mut token_identifier_words = [0u32; 8];
+    for (i, c) in address.0.chunks_exact(4).enumerate() {
+      token_identifier_words[i] = u32::from_le_bytes(c.try_into().unwrap());
+    }
+    athena_vm::syscalls::write_storage(&TOKEN_IDENTIFIER_KEY, &token_identifier_words);
+    address
```
noamnelke, line 42: Is the address only accessible during spawn? Wouldn't it be more efficient to read it from the host when needed instead of keeping it stored in the mutable state?

noamnelke (commented):
Generally looks good. I would just consider the two points I commented on. The wallet holding SMH is the more important one, imo.

examples/token/wallet/src/main.rs
```diff
@@ -0,0 +1,99 @@
...
+        .encode(),
+      ),
+      Some(MethodSelector::from("athexp_receive")),
+      0,
+    );
+  }
+
+  #[callable]
+  fn max_spend(&self) -> u64 {
+    // This wallet doesn't manage SMH and it cannot spend it.
```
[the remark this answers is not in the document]
poszu: Good point, I updated the code to have two methods: `spend()`, which works like a normal wallet spend (it sends SMH) and `send_token()`. The `max_spend()` now returns the SMH amount to send as it should.

> I would basically inherit all the normal functionality of a wallet and add a separate send_token() method.

Generally, yes, but I don't want to put too much into this example contract for two reasons:
1. I want to keep it simple to demonstrate the bare minimum needed for a token wallet (hence I skipped `proxy()` and `deploy()`)
2. I want to keep the functionality limited to keep the cost of maintaining it lower (less testing and SDK code needed etc.)

examples/token/mint/src/main.rs
poszu: Right. I updated the `Context` struct to keep the `callee` address as well and read it from there (instead of the mutable storage).

noamnelke (approved):
LGTM

